PR #1, opened 2016-03-09
Add Dockerfile and update README

naamunds:
cc: @MichaelSimons @dleeapho

### DIFF
--- a/preview/Dockerfile
+++ b/preview/Dockerfile
@@ -1,0 +1,15 @@
+FROM buildpack-deps:trusty-scm
+
+# Install the LKG preview build of the .NET CLI. This is done by downloading
+# and installing the .deb file directly because there is a significant latency
+# between the publishing of a .NET CLI Debian package and its availability on
+# the package repository, so the simple "apt-get install dotnet" might not
+# pick up the latest build.
+#
+# Because the dpkg command cannot install dependencies, treat errors as
+# warnings and then use apt-get to install the dependencies.
+RUN apt-get update \
+    && wget https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-ubuntu-x64.latest.deb \
+    && dpkg --force-depends -i dotnet-ubuntu-x64.latest.deb \
+    && apt-get -f -y --no-remove install \
+    && rm dotnet-ubuntu-x64.latest.deb
--- a/preview/README.md
+++ b/preview/README.md
@@ -1,0 +1,12 @@
+.NET CLI Preview Docker Image
+====================
+
+This repository contains the `Dockerfile` definition for Docker images that include the last-known-good (LKG) build of the [preview channel](https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/cli-installation-scenarios.md#channels) of [dotnet/cli]. See [dotnet/dotnet-docker](https://github.com/dotnet/dotnet-docker) for images with official releases of the .NET CLI.
+
+This project is part of .NET Core command-line (CLI) tools. You can find samples, documentation, and getting started instructions for .NET Core CLI tools at the [dotnet/cli] repo.
+
+# Supported tags and respective `Dockerfile` links
+
+-       [`preview`, `latest` (*preview/Dockerfile*)](https://github.com/dotnet/dotnet-docker-preview/blob/master/preview/Dockerfile)
+
+[dotnet/cli]: https://github.com/dotnet/cli
--- a/README.md
+++ b/README.md
@@ -1,2 +1,1 @@
-# dotnet-docker-preview
-Repo to power Docker images with preview bits of the .NET Core CLI.
+preview/README.md


colemickens:
How does this source: `https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-ubuntu-x64.latest.deb` compare with the releases in `https://apt-mo.trafficmanager.net/repos/dotnet`?

naamunds:
A couple implementation notes:
- We will switch to use the aka.ms link when it is available.
- The `dpkg` command cannot install dependencies. So, the Dockerfile treats errors from that command as warnings via the `--force-depends` option, and then invokes `apt-get -f -y install` to install the dependencies.

brthor:
:+1: 
@colemickens 
https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-ubuntu-x64.latest.deb

will currently map to the latest "dotnet" package available via the feed. Publishing to that feed has some lead time though so there will be a period where it's ahead.

colemickens:
Changing the download link from HTTPS to HTTP (aka.ms) seems like a bad idea (or I'm not sure how it adds any value).

colemickens:
@brthor Thanks for the info.

preview/README.md
```diff
@@ -0,0 +1,10 @@
+.NET CLI Preview Docker Image
```
MichaelSimons: What do you think about mentioning the other Dockerfile repository?  Something like 'See dotnet/dotnet-docker for officially released versions.'

preview/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM buildpack-deps:trusty-scm
+
+RUN apt-get update \
+    && wget https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-ubuntu-x64.latest.deb \
```
MichaelSimons: I think it is worth mentioning why the package is not retrieved from the Debian repository.  I think this may be a source of confusion as already observed in previous comments on this checkin.

preview/README.md
```diff
@@ -0,0 +1,10 @@
...
+====================
+
+This repository contains the `Dockerfile` definition for Docker images that include the latest build of the [preview channel](https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/cli-installation-scenarios.md#channels) of [dotnet/cli].
+
+This project is part of .NET Core command-line (CLI) tools. You can find samples, documentation, and getting started instructions for .NET Core CLI tools at the [dotnet/cli] repo.
+
+# Supported tags and respective `Dockerfile` links
+
+-       [`preview` (*preview/Dockerfile*)](https://github.com/dotnet/dotnet-docker-preview/blob/master/preview/Dockerfile)
```
MichaelSimons: This should also be tagged as 'latest'

preview/README.md
naamunds: That's a good point. I was also thinking that the title of the README in the dotnet-docker repo should be updated; currently it also says ".NET CLI Preview Docker Image".  I'm thinking just calling it ".NET CLI Docker Image" would be better.

preview/README.md
naamunds: Fixed.

preview/Dockerfile
naamunds: I've added a comment.

naamunds:
I've submitted another commit with some minor changes responding to comments here. I'll squash them into one commit before merging.

naamunds:
@colemickens The purpose of using aka.ms links is to add a layer of abstraction in the contract between .NET CLI and this Docker repo regarding the hosting of the CLI build artifacts. This way, if the CLI ever makes a breaking change to where artifacts are hosted, then the aka.ms links can be updated rather than the Dockerfile needing to be updated.

What are the risks of using an aka.ms URL that uses HTTP?  My understanding is that it would redirect to an HTTPS link, so I think the actual download would still use HTTPS.  I could be mistaken, though.

colemickens:
As a developer, I really, really don't want to see a link shortener in a place like this. The first thing people are going to do when they see the Dockerfile is resolve the link to see where they're randomly pulling DEB packages from to install. It just definitely feels off from what I would expect outside of MS projects. I don't know of any other projects that install DEBs from non-apt-repos, period, let alone via a non-https link shortener.

In my opinion, this is the wrong place to be worried about a contract. If there really needs to be a layer of indirection then it should be the use of a CNAME on top of the `apt-mo.trafficmanager.net` domain name. That way it can you could stop publishing to `apt-mo` entirely, publish to a new repo, and then update the CNAME to point at the new repo location.

If the primary apt repo stops being published to... a lot more than this one Dockerfile will be affected. Literally anyone who has installed `dotnet` via apt will just simply stop receiving updates... the CNAME would prevent this from ever being an issue.

colemickens:
(The risk is that someone can intercept the aka.ms http call and return a 302 to anywhere. Like a backdoored DEB, or something. Probably not a huge risk since these will be hopefully be built by some CI infrastructure and pushed to Docker Hub where most people will pull directly, but I still think that the CNAME is a much better choice for indirection/contract concerns).

friism:
Can't you just use a https url for `aka.ms`? It appear to support https, but maybe I'm missing something.

Agree that a http url (even if it's just a redirect) is very bad.

colemickens:
My mistake, last I'd heard it still didn't work with HTTPS properly. Still, this sort of redirection is not something done by other OSS projects and won't solve this general problem of a moving APT repo for people who are actually installing via apt on their development machines and want updates. Using a CNAME gets you everything aka.ms does, but solves the problem for more people.

brthor:
@colemickens The aka.ms link is not used to point to the apt repo but to the artifacts generated by the cli build which then get uploaded to the repo. The layout, location, and other details of those artifacts are implementation details of our build and this layer of indirection shields the dockerfile from those details.

We want to update this dockerfile immediately after the dotnet cli build finishes, and generates a new .deb package but our repo currently has a long (1 - ? hours) lead time after a package is added to it's queue.

For that reason we are using the direct link to the build artifact rather than the repo when this docker build is triggered.

MichaelSimons:
:shipit:

dleeapho:
:shipit:

dleeapho:
Re the aka.ms redirect. I agree that it isn't in keeping with the best practice that Dockerfiles should be as transparent as possible. How often is the CLI build artifact location going change anyway? If we need to change it in the Dockerfile, at least we are in control of that directly. I say we scrap the aka.ms redirection.

piotrpMSFT:
@dleeapho and I chatted offline. The target of the aka link changes with each build of the CLI repo.

PR #2, opened 2016-03-15
Temporarily work around LTTng assert

naamunds:
cc: @MichaelSimons @dleeapho

### DIFF
--- a/preview/Dockerfile
+++ b/preview/Dockerfile
@@ -1,5 +1,11 @@
 FROM buildpack-deps:trusty-scm
 
+# Work around https://github.com/dotnet/cli/issues/1582 until Docker releases a
+# fix (https://github.com/docker/docker/issues/20818). This workaround allows
+# the container to be run with the default seccomp Docker settings by avoiding
+# the restart_syscall made by LTTng which causes a failed assertion.
+ENV LTTNG_UST_REGISTER_TIMEOUT 0
+
 # Install the LKG preview build of the .NET CLI. This is done by downloading
 # and installing the .deb file directly because there is a significant latency
 # between the publishing of a .NET CLI Debian package and its availability on


preview/Dockerfile
```diff
@@ -1,5 +1,16 @@
...
 
+# Temporarily work around https://github.com/dotnet/cli/issues/1582 by
+# avoiding the restart_syscall made by LTTng which causes a failed assertion.
+# This workaround specifies that applications should not wait for a sessiond
+# "registration done" command before proceeding to execute the main program.
+#
+# This issue is being addressed in Docker (see
+# https://github.com/docker/docker/issues/20818). When that fix has been
+# included in a release of Docker, this workaround will be removed, and users
+# will be expected to upgrade their Docker installation.
```
MichaelSimons: In a way I think this is a little verbose but that is just my opinion.  One thing that is incorrect is to say users are "expected to upgrade".  Users can use the seccomp functionality to either run in unconfined mode or with a custom profile that allows the restart_syscall.  The reason we are doing this workaround is to simply provide the best out of box experience for those who want to kick the tires.

I think you could get away with simply saying "Workaround https://github.com/dotnet/cli/issues/1582 until Docker releases a fix (https://github.com/docker/docker/issues/20818).  This workaround allows the container to be run with the default seccomp docker settings."

naamunds: That makes sense.  I've updated the comment to mostly match your suggestion, but I do like mentioning that it avoids the failed assertion from LTTng.

preview/Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM buildpack-deps:trusty-scm
 
+# Work around dotnet/cli#1582 until Docker releases a fix (docker/docker#20818).
```
MichaelSimons: I didn't intend to use the abbreviated github link syntax.  I think github shortened what I had typed into my comment.  I think the Dockerfile should use the real link.

naamunds: OK, I was wondering about that. I figured you did it intentionally.

I'll make that change to use the full URL. Other than that, does it look good now?

MichaelSimons: Yes it does.

dleeapho:
LGTM

PR #6, opened 2016-03-23
Added logic to cleanup the apt cache.

### LINKED ISSUE #5: Images are too large (and include apt 's cache)
I'd recommend clearing the apt cache in the same RUN invocation that you update it in. It ensures that you don't wind up making everyone download a copy of your apt cache.

example:
https://github.com/colemickens/polykube/blob/637b17dec481e8fba86d0628c13469efd019d485/dotnet/dockerfiles/dotnet-base/Dockerfile#L10

MichaelSimons:
This reduces the image size by ~21mb.  This resolves #5.

@naamunds

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/dotnet-docker-preview/6)

### DIFF
--- a/preview/Dockerfile
+++ b/preview/Dockerfile
@@ -18,4 +18,5 @@ RUN apt-get update \
     && wget https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-ubuntu-x64.latest.deb \
     && dpkg --force-depends -i dotnet-ubuntu-x64.latest.deb \
     && apt-get -f -y --no-remove install \
-    && rm dotnet-ubuntu-x64.latest.deb
+    && rm dotnet-ubuntu-x64.latest.deb \
+    && rm -rf /var/lib/apt/lists/*


brthor:
I didn't realize the apt cache was so large!

preview/Dockerfile
```diff
@@ -18,4 +18,5 @@ RUN apt-get update \
     && wget https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-ubuntu-x64.latest.deb \
     && dpkg --force-depends -i dotnet-ubuntu-x64.latest.deb \
     && apt-get -f -y --no-remove install \
-    && rm dotnet-ubuntu-x64.latest.deb
+    && rm dotnet-ubuntu-x64.latest.deb \
+    && rm -rf /var/lib/apt/lists/*
```
naamunds: What do you think about also calling `apt-get clean`?  I'm wondering if the [Dockerfile best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get) forgot to include that in their example, because it seems like their explanation of the code implies it was included (or that something was used beyond deleting /var/lib/apt/lists/* at least) when it says "In addition, cleaning up the apt cache and removing /var/lib/apt/lists helps keep the image size down."

colemickens: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get

That says that `apt-get clean` is run automatically.

naamunds: Ah, I missed that note. Thanks.

naamunds:
LGTM

PR #9, opened 2016-04-18
Update Dockerfile to be Debian based

MichaelSimons:
All of our Docker images will be Debian based for RC2 (https://github.com/dotnet/dotnet-docker/issues/11).  CLI now supports Debian therefore the preview image was updated accordingly.  Keep in mind the CLI Debian installer work is out of scope for RC2 therefore the Dockerfile needs to contain the logic to install CLI.

/cc @naamunds, @dleeapho

### DIFF
--- a/preview/Dockerfile
+++ b/preview/Dockerfile
@@ -8,7 +8,7 @@ ENV LTTNG_UST_REGISTER_TIMEOUT 0
 
 # Install .NET CLI dependencies
 RUN echo "deb [arch=amd64] http://llvm.org/apt/jessie/ llvm-toolchain-jessie-3.6 main" > /etc/apt/sources.list.d/llvm.list \
-    && wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key|apt-key add - \
+    && wget -q -O - http://llvm.org/apt/llvm-snapshot.gpg.key|apt-key add - \
     && apt-get update \
     && apt-get install -y --no-install-recommends \
         clang-3.5 \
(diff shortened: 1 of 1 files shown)


friism:
@MichaelSimons some unsolicited input:

 1. It'd be nice if you published and checked checksums for the `.tar.gz`
 2. Consider passing `-q` to both `wget` to make it less noisy
 3. Consider passing `-p` to `mkdir` (that create the full hierachy if it doesn't exist - it's not required here, but makes things more robust)
 4. Consider dropping `v` from the tar command to make it less noisy
 5. This order feels more natural to me:

```
    && rm dotnet-dev-debian-x64.latest.tar.gz \
    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
```

colemickens:
Just out of pure curiosity, is debian "better supported" than ubuntu, or what motivates the change here? (Also so I can make appropriate changes to my own Dockerfiles.)

MichaelSimons:
@colemickens both Ubuntu and Debian are officially supported platforms.  The [Dockerfile best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/) recommends using Debian as the basis for images since it’s very tightly controlled and kept minimal.  Lastly I have read that there are some [copyright concerns](http://mjg59.dreamwidth.org/36312.html) regarding Ubuntu images.

MichaelSimons:
@friism thanks for the feedback.  I made all of the changes you suggested except for the checksum.  I opened a [separate issue](https://github.com/dotnet/dotnet-docker-preview/issues/10) for this item and will address it with a separate PR.

preview/Dockerfile
```diff
@@ -1,28 +1,34 @@
-FROM buildpack-deps:trusty-scm
+FROM buildpack-deps:jessie-scm
```
naamunds: Have we considered using `jessie-curl` instead?  That image appears to be 42 MB smaller.  According to the [buildpack-deps repo](https://hub.docker.com/r/library/buildpack-deps/), the scm images add "various source control management tools"; I'm wondering how important those tools are for the base .NET Docker image.

friism: @naamunds +1 - you can also just install curl with --no-depends and not use `wget`

preview/Dockerfile
```diff
@@ -1,28 +1,34 @@
...
+        liblttng-ust0 \
+        libssl1.0.0 \
+        libstdc++6 \
+        libtinfo5 \
+        libunwind8 \
+        libuuid1 \
+        zlib1g \
     && rm -rf /var/lib/apt/lists/*
+
+# Install .NET CLI
```
naamunds: It might be useful to still mention that we are installing the LKG preview build of the .NET CLI. Also, what do you think about mentioning that we are extracting the tarball directly because an installer for Debian is not yet available?

naamunds:
LGTM, with a couple minor comments.

preview/Dockerfile
MichaelSimons: @naamunds, Can you open a separate issue for this?  It keeps popping up and I think it is best to capture the discussion in an issue versus this PR.  I think this PR is should be simply about switching from Ubuntu to Debian.

preview/Dockerfile
MichaelSimons: I changed the comment to `Install the LKG preview build of the .NET CLI.'  After this PR is merged, I am going to log a new issue to update the Dockerfile to use the Debian package once https://github.com/dotnet/cli/issues/2565 is fixed.  IMO it is not necessary to comment on this within the Dockerfile.

dleeapho:
:shipit:

preview/Dockerfile
naamunds: I've logged https://github.com/dotnet/dotnet-docker-preview/issues/11 for this.

PR #21, opened 2016-06-21
Triggered the building of local package cache

### LINKED ISSUE #20: Dockerfile should trigger the building of the local package cache.
With the introduction of the [offline feature](https://github.com/dotnet/cli/pull/3485), the dotnet CLI builds a local package cache on first use.  The dotnet Dockerfile should trigger this to happen.  This takes some time therefore it will be a savings for all users of the image.  Additionally having the package cache in the base image will save disc space for scenarios in which base image is used multiple times on the same machine as the base layer will be shared.

MichaelSimons:
Fixes #20 

@naamunds, @dleeapho

### DIFF
--- a/preview/Dockerfile
+++ b/preview/Dockerfile
@@ -28,3 +28,11 @@ RUN wget -q https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/Late
     && tar -zxf dotnet-dev-debian-x64.latest.tar.gz -C /usr/share/dotnet \
     && rm dotnet-dev-debian-x64.latest.tar.gz \
     && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+
+# Trigger the population of the local package cache
+ENV NUGET_XMLDOC_MODE skip
+RUN mkdir warmup \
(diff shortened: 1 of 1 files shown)


preview/Dockerfile
```diff
@@ -28,3 +28,11 @@ RUN wget -q https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/Late
...
     && rm dotnet-dev-debian-x64.latest.tar.gz \
     && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+
+# Trigger the population of the local package cache
+ENV NUGET_XMLDOC_MODE skip
+RUN mkdir warmup \
+    && cd warmup \
+    && dotnet new \
+    && cd .. \
+    && rm -r warmup
```
naamunds: Would it be safer to add the -f flag (`rm -rf warmup`), in case it prompts otherwise?

naamunds:
LGTM

preview/Dockerfile
MichaelSimons: Changes as suggested.

Issue #22, opened 2016-06-28
The repository should be updated to pick up the next-release train of the CLI bits


dleeapho:

MichaelSimons:
Along with updating the Dockerfile, the [Maestro hook](https://github.com/dotnet/versions/blob/master/Maestro/subscriptions.json#L151) will need to be updated so that the Docker Hub build gets triggered when a new CLI build is completed for the 'next-release'.

naamunds:
Leaving open until the Maestro hook is updated.

Issue #14, opened 2016-05-12
We should rename this repo and the Docker Hub repository to something other than "preview"


dleeapho:
Originally we had named this image dotnet-preview to match the fact that the image uses the .NET CLI Preview channel but with the advent of "preview1" as a version name in the CLI we risk confusion. We should find a better name-- something like "nightly". @blackdwarf @glennc do you have suggestions.

glennc:
I am ok with Nightly. @blackdwarf what do you guys refer to the builds of the CLI on GitHub? Do you have a name?

naamunds:
I synced up with @blackdwarf over email, and he agrees with "nightly".  Also, the image tag will be named "rel-1.0.0" to align with the CLI channel/branch name.

The command to pull the image will be:
`docker pull microsoft/dotnet-nightly:rel-1.0.0`

PR #23, opened 2016-06-30
  Update image to use rel/1.0.0 builds 

naamunds:
The rel/1.0.0 branch represents the next release of the .NET CLI.

Fixes https://github.com/dotnet/dotnet-docker-nightly/issues/22.  Also related to https://github.com/dotnet/dotnet-docker-nightly/issues/14 (rename from "preview" to "nightly").

cc: @MichaelSimons @dleeapho


naamunds:
Unfortunately it looks like the history for the README.md can't be preserved when viewed in GitHub after the move from the `preview` directory to the `rel-1.0.0` directory.  The history can still be seen from the command line by using `git log --follow rel-1.0.0/README.md`, but it looks like GitHub still doesn't support tracking that history even though it's been requested for a while (see http://stackoverflow.com/questions/5646174/github-follow-history-by-default).

You can see the diff of the file by viewing https://github.com/dotnet/dotnet-docker-nightly/commit/4246659a3f8f30d0a940a1da0653b490249cba67.

rel-1.0.0/Dockerfile
```diff
@@ -22,8 +22,8 @@ RUN apt-get update \
         zlib1g \
     && rm -rf /var/lib/apt/lists/*
 
-# Install the LKG preview build of the .NET CLI
-RUN wget -q https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/Latest/dotnet-dev-debian-x64.latest.tar.gz \
+# Install the LKG rel/1.0.0 build of the .NET CLI
```
MichaelSimons: What do you think of removing this from the comment?  IMO it doesn't add value and is a maintenance issue.

MichaelSimons:
:shipit:

rel-1.0.0/Dockerfile
naamunds: Makes sense.  I've made that change.

PR #25, opened 2016-08-02
Add new image types to match official repo

naamunds:
cc: @MichaelSimons

### DIFF
--- a/rel-1.0.0/README.md
+++ b/rel-1.0.0/README.md
@@ -16,24 +16,24 @@ This project is part of the .NET Core CLI tools. You can find samples, documenta
 ## Supported tags
 
 ### Development images
--       [`sdk`, `latest` (*rel-1.0.0/debian/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/rel-1.0.0/debian/Dockerfile)
--       [`windowsservercore` (*rel-1.0.0/windowsservercore/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/rel-1.0.0/windowsservercore/Dockerfile)
--       [`nanoserver` (*rel-1.0.0/nanoserver/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/rel-1.0.0/nanoserver/Dockerfile)
--       [`onbuild` (*rel-1.0.0/onbuild/Dockerfile*)](https://github.com/dotnet/dotnet-docker/blob/master/rel-1.0.0/debian/onbuild/Dockerfile)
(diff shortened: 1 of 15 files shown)


rel-1.0.0/windowsservercore/onbuild/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM microsoft/dotnet-nightly:windowsservercore
```
MichaelSimons: I think all of the references to images from the dotnet-nightly repo should be using the fully qualified tag names (e.g. rel-1.0.0-windowsservercore) so that they don't break as the shortened tags are moved to other images over time.

rel-1.0.0/README.md
```diff
@@ -1,14 +1,88 @@
...
 See [dotnet/dotnet-docker](https://github.com/dotnet/dotnet-docker) for images with official releases of [.NET Core](http://dotnet.github.io).
 
 This project is part of the .NET Core CLI tools. You can find samples, documentation, and getting started instructions at the [dotnet/cli] repo.
 
-# Supported tags and respective `Dockerfile` links
+[![Downloads from Docker Hub](https://img.shields.io/docker/pulls/microsoft/dotnet-nightly.svg)](https://registry.hub.docker.com/u/microsoft/dotnet-nightly)
+[![Stars on Docker Hub](https://img.shields.io/docker/stars/microsoft/dotnet-nightly.svg)](https://registry.hub.docker.com/u/microsoft/dotnet-nightly)
+
+
+## Supported tags
```
MichaelSimons: >Supported tags [](start = 3, length = 14)

I think fully versioned tags need to exist for each of the images.  These are the tags that should be used in the onbuild, core images when referencing the base image.

rel-1.0.0/README.md
```diff
@@ -1,14 +1,88 @@
...
+
+### Development images
+-       [`sdk`, `latest` (*rel-1.0.0/debian/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/rel-1.0.0/debian/Dockerfile)
```
MichaelSimons: >sdk [](start = 10, length = 3)

I think this should be replaced by the full version tag.

rel-1.0.0/README.md
naamunds: Yeah, that is preferable for maintenance reasons.  I've made the change.

rel-1.0.0/README.md
naamunds: Fixed.

rel-1.0.0/windowsservercore/onbuild/Dockerfile
naamunds: Fixed.

MichaelSimons:
:shipit:

PR #26, opened 2016-08-09
Add app to automatically update dependency versions

naamunds:
The update-dependencies .NET app must be invoked from the root of the
repo.  A update-dependencies.ps1 script is also provided which installs
the .NET CLI and then compiles and invokes the app; this script can be
invoked from anywhere.

The plan is for Maestro to invoke a VSTS build definition whenever a new
CLI build is available, which will invoke the update-dependencies.ps1
script.

Also add a .gitignore file.

cc: @MichaelSimons

### DIFF
--- a/update-dependencies/Config.cs
+++ b/update-dependencies/Config.cs
@@ -1,0 +1,53 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dotnet.Docker.Nightly
+{
+    public class Config
+    {
+        public static Config s_Instance { get; } = new Config();
+
+        private Lazy<string> _userName = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_USER"));
+        private Lazy<string> _email = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_EMAIL"));
+        private Lazy<string> _password = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_PASSWORD"));
+        
+        private Lazy<string> _cliVersionUrl = new Lazy<string>(() => $"https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/cli/{GetEnvironmentVariable("CLI_BRANCH")}");
+        private Lazy<string> _gitHubUpstreamOwner = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_UPSTREAM_OWNER", "dotnet"));
+        private Lazy<string> _gitHubProject = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_PROJECT", "dotnet-docker-nightly"));
+        private Lazy<string> _gitHubUpstreamBranch = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_UPSTREAM_BRANCH", "master"));
+        private Lazy<string[]> _gitHubPullRequestNotifications = new Lazy<string[]>(() => 
+                                                GetEnvironmentVariable("GITHUB_PULL_REQUEST_NOTIFICATIONS", "")
+                                                    .Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries));
+
+        private Config()
+        {
+        }
+
+        public string UserName => _userName.Value;
+        public string Email => _email.Value; 
+        public string Password => _password.Value;
+        public string CliVersionUrl => _cliVersionUrl.Value;
+        public string GitHubUpstreamOwner => _gitHubUpstreamOwner.Value;
+        public string GitHubProject => _gitHubProject.Value;
+        public string GitHubUpstreamBranch => _gitHubUpstreamBranch.Value;
+        public string[] GitHubPullRequestNotifications => _gitHubPullRequestNotifications.Value;
+
+        private static string GetEnvironmentVariable(string name, string defaultValue = null)
+        {
+            string value = Environment.GetEnvironmentVariable(name);
+            if (value == null)
+            {
+                value = defaultValue;
+            }
+
+            if (value == null)
+            {
+                throw new InvalidOperationException($"Can't find environment variable '{name}'.");
+            }
+
+            return value;
+        }
+    }
+}
--- a/update-dependencies/Program.cs
+++ b/update-dependencies/Program.cs
@@ -1,0 +1,113 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.VersionTools;
+using Microsoft.DotNet.VersionTools.Automation;
+using Microsoft.DotNet.VersionTools.Dependencies;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Dotnet.Docker.Nightly
+{
+    public static class Program
+    {
+        private static readonly string s_repoRoot = Directory.GetCurrentDirectory();
+        private static readonly Config s_config = Config.s_Instance;
+        private static bool s_updateOnly = false;
+
+        public static void Main(string[] args)
+        {
+            if (ParseArgs(args))
+            {
+                DependencyUpdateResults updateResults = UpdateFiles();
+
+                if (!s_updateOnly && updateResults.ChangesDetected())
+                {
+                    CreatePullRequest(updateResults).Wait();
+                }
+            }
+        }
+        
+        private static bool ParseArgs(string[] args)
+        {
+            foreach (string arg in args)
+            {
+                if (string.Equals(arg, "--Update", StringComparison.OrdinalIgnoreCase))
+                {
+                    s_updateOnly = true;
+                }
+                else if (arg.Contains('='))
+                {
+                    int delimiterIndex = arg.IndexOf('=');
+                    string name = arg.Substring(0, delimiterIndex);
+                    string value = arg.Substring(delimiterIndex + 1);
+
+                    Environment.SetEnvironmentVariable(name, value);
+                }
+                else
+                {
+                    Console.Error.WriteLine($"Unrecognized argument '{arg}'");
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private static DependencyUpdateResults UpdateFiles()
+        {
+            // The BuildInfo does not contain the CLI product version, so the version is retrieved from a
+            // particular CLI package (e.g. Microsoft.DotNet.Cli.Utils). Once the BuildInfo includes the
+            // product version, it should be utilized.
+            //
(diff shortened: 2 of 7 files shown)


naamunds:
Here is a sample PR created with this app, using my fork of the repo:
https://github.com/naamunds/dotnet-docker-nightly/pull/1

update-dependencies/update-dependencies.ps1
```diff
@@ -0,0 +1,71 @@
...
+
+$Architecture='x64'
+
+$RepoRoot = "$PSScriptRoot\.."
+$AppPath = "$PSScriptRoot"
+
+# Use a repo-local install directory (but not the artifacts directory because that gets cleaned a lot)
+if (!$env:DOTNET_INSTALL_DIR)
+{
+    $env:DOTNET_INSTALL_DIR="$RepoRoot\.dotnet_stage0\Windows\$Architecture"
```
MichaelSimons: >.dotnet_stage0 [](start = 39, length = 14)

What do you think about naming this something besides stage0?  This IMO is something that makes sense in CLI but not in this context.  I would lean towards simple .dotnet

update-dependencies/update-dependencies.ps1
```diff
@@ -0,0 +1,71 @@
...
+    mkdir $env:DOTNET_INSTALL_DIR | Out-Null
+}
+
+if (!(Test-Path "$RepoRoot\artifacts"))
+{
+    mkdir "$RepoRoot\artifacts" | Out-Null
+}
+
+# Install a stage 0
+$DOTNET_INSTALL_SCRIPT_URL="https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1"
```
MichaelSimons: >rel/1.0.0 [](start = 73, length = 9)

I tend to think this should be parameterized with possibly this being the default.

update-dependencies/update-dependencies.ps1
```diff
@@ -0,0 +1,71 @@
...
+Invoke-WebRequest $DOTNET_INSTALL_SCRIPT_URL -OutFile "$RepoRoot\artifacts\dotnet-install.ps1"
+
+& "$RepoRoot\artifacts\dotnet-install.ps1" -Architecture $Architecture
+if($LASTEXITCODE -ne 0) { throw "Failed to install stage0" }
+
+# Put the stage0 on the path
+$env:PATH = "$env:DOTNET_INSTALL_DIR;$env:PATH"
```
MichaelSimons: This doesn't seem necessary as it should be done by dotnet-install.ps1.

update-dependencies/update-dependencies.ps1
```diff
@@ -0,0 +1,71 @@
...
+if (!(Test-Path "$RepoRoot\artifacts"))
+{
+    mkdir "$RepoRoot\artifacts" | Out-Null
+}
+
+# Install a stage 0
+$DOTNET_INSTALL_SCRIPT_URL="https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1"
+Invoke-WebRequest $DOTNET_INSTALL_SCRIPT_URL -OutFile "$RepoRoot\artifacts\dotnet-install.ps1"
+
+& "$RepoRoot\artifacts\dotnet-install.ps1" -Architecture $Architecture
```
MichaelSimons: What do you think about explicitly passing in DOTNET_INSTALL_DIR as the InstallDir argument instead of relying on the implementation of the script reading the environment variable.  IMO the benefits of this is it makes the code more readable and eliminates more chances of the script breaking.

update-dependencies/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+            );
+
+            return prCreator.CreateOrUpdateAsync(commitMessage, commitMessage, string.Empty);
+        }
+
+        private static IEnumerable<IDependencyUpdater> GetUpdaters()
+        {
+            return Directory.GetFiles(s_repoRoot, "Dockerfile", SearchOption.AllDirectories)
+                .Where(path => File.ReadAllText(path).Contains("ENV DOTNET_SDK_VERSION")) 
+                .Select(path => CreateRegexUpdater(path, "Microsoft.DotNet.Cli.Utils"));
```
MichaelSimons: >CreateRegexUpdater(path, "Microsoft.DotNet.Cli.Utils") [](start = 32, length = 54)

I think it might be worthwhile to have a code comment explaining why "Microsoft.DotNet.Cli.Utils" is used versus reading some more abstract information about the CLI.

update-dependencies/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+            IEnumerable<BuildInfo> buildInfos = new[] { BuildInfo.Get("Cli", s_config.CliVersionUrl, fetchLatestReleaseFile: false) };
+            IEnumerable<IDependencyUpdater> updaters = GetUpdaters();
+
+            DependencyUpdater updater = new DependencyUpdater();
+            return updater.Update(updaters, buildInfos);
+        }
+
+        private static Task CreatePullRequest(DependencyUpdateResults updateResults)
+        {
+            string commitMessage = $"Update CLI to {updateResults.UsedBuildInfos.Single().LatestReleaseVersion}";
```
MichaelSimons: >CLI [](start = 44, length = 3)

What do you think about using the term SDK instead of CLI?  In general the Dockerfiles use the SDK nomenclature.

update-dependencies/Config.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public class Config
+    {
+        public static Config s_Instance { get; } = new Config();
+
+        private Lazy<string> _userName = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_USER"));
+        private Lazy<string> _email = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_EMAIL"));
+        private Lazy<string> _password = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_PASSWORD"));
+        
+        private Lazy<string> _cliVersionUrl = new Lazy<string>(() => GetEnvironmentVariable("CLI_VERSION_URL", "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/cli/rel/1.0.0"));
```
MichaelSimons: >"https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/cli/rel/1.0.0" [](start = 111, length = 90)

Is this default useful?  I ask this because the value appears to only be applicable for v1 instead of the long term.  I would not recommend utilizing this default in our usage therefore I wonder if it is valuable.

update-dependencies/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                s_config.GitHubProject,
+                s_config.GitHubUpstreamOwner,
+                s_config.GitHubUpstreamBranch,
+                s_config.UserName
+            );
+
+            return prCreator.CreateOrUpdateAsync(commitMessage, commitMessage, string.Empty);
+        }
+
+        private static IEnumerable<IDependencyUpdater> GetUpdaters()
```
MichaelSimons: >GetUpdaters [](start = 55, length = 11)

I think it would be worthwhile to note why an updater only exists for the SDK and not the runtime/host.

update-dependencies/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            return Directory.GetFiles(s_repoRoot, "Dockerfile", SearchOption.AllDirectories)
+                .Where(path => File.ReadAllText(path).Contains("ENV DOTNET_SDK_VERSION"))
```
MichaelSimons: >.Where(path => File.ReadAllText(path).Contains("ENV DOTNET_SDK_VERSION"))  [](start = 16, length = 74)

Is this unnecessary overhead?   Why not just create the package updater for all Dockerfiles?  This implementation seems like it is going to cause Dockerfiles to be read multiple times.

update-dependencies/update-dependencies.ps1
naamunds: Good point.  I see that it also creates the InstallDir if it doesn't exist, so I've removed that logic from this script also.

update-dependencies/update-dependencies.ps1
naamunds: I've renamed the default install directory to `.dotnet`.

update-dependencies/update-dependencies.ps1
naamunds: Good idea.

update-dependencies/update-dependencies.ps1
naamunds: Makes sense. I've added a DotnetInstallDir parameter to this script, which defaults to `"$RepoRoot\.dotnet\Windows\$Architecture"`.  The call to dotnet-install.ps1 now uses $DotnetInstallDir for the InstallDir parameter.

update-dependencies/Config.cs
naamunds: Good point; the CLI version should always be explicitly passed.  What do you think about instead using a CLI_BRANCH environment variable, and then the CLI_VERSION_URL would be computed from it as follows:
`private Lazy<string> _cliVersionUrl = new Lazy<string>(() => string.Format("https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/cli/{0}", GetEnvironmentVariable("CLI_BRANCH")));`

update-dependencies/Program.cs
naamunds: Updated as suggested.

update-dependencies/Program.cs
naamunds: I did this to avoid the overhead of having additional IDependencyUpdaters, but I think you're right that having additional IDependencyUpdaters is preferable to reading some of the Dockerfiles multiple times.  I've removed this line as suggested.

update-dependencies/Program.cs
naamunds: I've added a comment for this.

update-dependencies/Program.cs
naamunds: Fixed.

update-dependencies/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            return true;
+        }
+
+        private static DependencyUpdateResults UpdateFiles()
+        {
+            // Update the version of the .NET Core SDK in the Dockerfiles to the latest version in the
+            // specified CLI channel. This version is determined by using DotNet.VersionTools to examine the
+            // latest version of one of the packages produced by the CLI build (Microsoft.DotNet.Cli.Utils),
+            // which has the same version as the SDK itself.
```
MichaelSimons: My intent was not to reiterate what the code does.  I think the code is readable already.  My point was to simply explain why the code was grabbing the Microsoft.DotNet.Cli.Utils version versus the CLI "Product".  

"The BuildInfo does not contain the CLI product version therefore the version is retrieved from a particular CLI package (eg. Microsoft.DotNet.Cli.Utils).  Once the BuildInfo includes the product version, it should be utilized."

MichaelSimons:
One minor comment, other than that it looks good.

update-dependencies/Program.cs
naamunds: Updated as suggested.

PR #30, opened 2016-08-10
Fix regex to handle CRLF line endings

naamunds:
The regex pattern for the DOTNET_SDK_VERSION was including \r in its
matches, causing the \r to be removed from that line when the version
string was replaced. This is because by default the '.' language element
matches every character except \n. The fix is to explicitly match every
character except \r and \n so that the pattern matches don't include \r.

This change also adds a Trace Listener so that useful output from
VersionTools is captured in the output of the app.

cc: @MichaelSimons @dagood


MichaelSimons:
:shipit:

PR #35, opened 2016-08-19
Add Jenkins CI and automated tests for Debian

naamunds:
Whenever a PR is created in the dev or master branch, a Jenkins job will
build each Debian Dockerfile and run tests.  The tests perform a simple
smoke test on the images.

In this initial implementation, only the Debian images are built and
tested.  The plan is to eventually test the Windows Server images also.

/cc: @MichaelSimons @mmitche

### DIFF
--- a/build-and-test.sh
+++ b/build-and-test.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -e 	# Exit immediately upon failure
+set -o pipefail  # Carry failures over pipes
+
+repo_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+docker_repo="microsoft/dotnet-nightly"
+
+function build_dockerfiles {
+    for dockerfile_dir in $( egrep -v 'nanoserver|windowsservercore' <<< "${1}" | sort ); do
+        tag="${docker_repo}:$( sed -e 's/.\///' -e 's/debian\///' -e 's/debian/sdk/' -e 's/\//-/g' <<< "${dockerfile_dir}" )"
+        echo "----- Building ${tag} -----"
+        docker build -t "${tag}" "${dockerfile_dir}"
+    done
+}
+
+pushd "${repo_root}" > /dev/null
+
+build_dockerfiles "$( find . -path './.*' -prune -o -name 'Dockerfile' -a -path '*/*-deps/*' -print0 | xargs -0 -n1 dirname )"
+
+build_dockerfiles "$( find . -path './.*' -prune -o -name 'Dockerfile' -a ! -path '*/*-deps/*' -print0 | xargs -0 -n1 dirname)"
+
+popd > /dev/null
+
+"${repo_root}/test/run-test.sh"
--- a/netci.groovy
+++ b/netci.groovy
@@ -1,0 +1,17 @@
+import jobs.generation.Utilities
+
+def project = GithubProject
+def branch = GithubBranchName
+def isPr = true
+def containerOs = 'Debian'
+def newJobName = Utilities.getFullJobName(project, containerOs, isPr)
+
+def newJob = job(newJobName) {
+    steps {
+        shell(".build-and-test.sh")
+    }
+}
+
+Utilities.setMachineAffinity(newJob, 'Ubuntu16.04', 'latest-or-auto-docker')
+Utilities.standardJobSetup(newJob, project, isPr, "*/${branch}")
+Utilities.addGithubPRTriggerForBranch(newJob, branch, containerOs)
--- a/test/run-test.sh
+++ b/test/run-test.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+set -e 	# Exit immediately upon failure
+set -o pipefail  # Carry failures over pipes
+
+repo_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/.."
+docker_repo="microsoft/dotnet-nightly"
+
+if [ -z "${DEBUGTEST}" ]; then
+    optional_docker_run_args="--rm"
+fi
+
+pushd "${repo_root}" > /dev/null
+
+for cli_channel in $( find . -mindepth 1 -maxdepth 1 -type d ! -name '.*' ! -name 'test' ! -name 'update-dependencies' -print | sed -e 's/\.\///' ); do
+    tag_base="${docker_repo}:${cli_channel}"
+
+    app_name="app$(date +%s)"
+    app_dir="${repo_root}/.test-assets/${app_name}"
+    mkdir -p "${app_dir}"
+
+    echo "----- Testing ${tag_base}-sdk -----"
+    docker run -t "${optional_docker_run_args}" -v "${app_dir}:/${app_name}" -v "${repo_root}/test:/test" --name "sdk-test-${app_name}" --entrypoint /test/create-run-publish-app.sh "${tag_base}-sdk" "${app_name}"
+
+    echo "----- Testing ${tag_base}-core -----"
+    docker run -t "${optional_docker_run_args}" -v "${app_dir}:/${app_name}" --name "core-test-${app_name}" --entrypoint dotnet "${tag_base}-core" "/${app_name}/publish/${app_name}.dll"
+
+    echo "----- Testing ${tag_base}-onbuild -----"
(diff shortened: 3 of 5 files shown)


naamunds:
test ci please

test/build-images.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e 	# Exit immediately upon failure
+
+: ${1?"Need to pass the CLI channel for which to build the Docker images"}
```
MichaelSimons: Can't this be discovered from within this script so that as channels are added, the netci.groovy script doesn't have to be updated each time?

test/build-images.sh
```diff
@@ -0,0 +1,28 @@
...
+docker build -t ${tag_base}-sdk ${image_dir_base}
+
+echo "----- Building ${tag_base}-core -----"
+docker build -t ${tag_base}-core ${image_dir_base}/core
+
+echo "----- Building ${tag_base}-core-deps -----"
+docker build -t ${tag_base}-core-deps ${image_dir_base}/core-deps
+
+echo "----- Building ${tag_base}-onbuild -----"
+docker build -t ${tag_base}-onbuild ${image_dir_base}/onbuild
```
MichaelSimons: I was thinking that the script would self discover the dockerfiles, build them, and then infer the tag name from the directory structure.  This will be less maintenance overtime as tags are added/removed.

mmitche:
LGTM on netci file.

mmitche:
@dotnet-bot test ci please

netci.groovy
```diff
@@ -0,0 +1,19 @@
...
+def branch = GithubBranchName
+def isPr = true
+def containerOs = 'Debian' 
+def cliChannel = 'rel-1.0.0'
+def newJobName = Utilities.getFullJobName(project, containerOs, isPr)
+
+def newJob = job(newJobName) {
+    steps {
+        shell("./test/build-images.sh ${cliChannel}")
+        shell("./test/run-test.sh ${cliChannel}")
```
MichaelSimons: Shouldn't this be updated to invoke `build-and-test.sh` and also cleanup the cliChannel?

build-and-test.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e 	# Exit immediately upon failure
+set -o pipefail  # Carry failures over pipes
+
+repo_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+docker_repo="microsoft/dotnet-nightly"
+
+pushd "${repo_root}" > /dev/null
+
+find . -path './.git' -prune -o -path './test' -prune -o -path './update-dependencies' -prune -o -name 'Dockerfile' -printf '%p\t%d\n' | egrep -v 'nanoserver|windowsservercore' | sort -n -k2 | while read line
```
MichaelSimons: >-path './.git' -prune -o -path './test' -prune -o -path './update-dependencies' -prune -o -name 'Dockerfile' [](start = 7, length = 108)

Do you think it is strictly necessary to exclude all of these paths?  They don't have Dockerfiles.  I would prefer to avoid the special rules that need to be maintained as additional dirs are added/removed.

MichaelSimons: >-printf [](start = 116, length = 7)

Does this work on OSX?  Alternatively you could do something like the following which I think will work on OSX.  This approach avoids the need to track the directory depth.

find . -name Dockerfile -print0 | xargs -0 -n1 dirname | egrep -v 'nanoserver|windowsservercore' | sort | while read line
do
    dockerfile_dir=${line}
    tag="${docker_repo}:$( sed -e 's/\.\///' -e 's/debian\///' -e 's/debian/sdk/' -e 's/\//-/g' <<< "${dockerfile_dir}" )"
    echo "----- Building ${tag} -----"
    docker build -t "${tag}" "${dockerfile_dir}"
done

naamunds: The `test` folder may have Dockerfiles in the `artifacts` subfolder created by `run-test.sh`, so we should at least exclude that one. I think we can safely assume that the `.git` folder won't have Dockerfiles, but I think that the `update-dependencies` folder could in the future, so I decided to just exclude them all.  I can remove the exclusion of the `.git` folder, but it seems like it might be good to still explicitly exclude the `update-dependencies` folder.  What are your thoughts on that?

MichaelSimons: Ah yes - what do you think about moving the artifacts folder used by test to the root and then excluding just that single folder?

---
In reply to: [76320309](https://github.com/dotnet/dotnet-docker-nightly/pull/35#discussion_r76320309) [](ancestors = 76320309)

test/run-test.sh
```diff
@@ -0,0 +1,35 @@
...
+set -o pipefail  # Carry failures over pipes
+
+repo_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/.."
+docker_repo="microsoft/dotnet-nightly"
+
+if [ -z "${DEBUGTEST}" ]; then
+    optional_docker_run_args="--rm"
+fi
+
+for cli_channel in $( find . -mindepth 1 -maxdepth 1 -type d ! -name .git ! -name test ! -name update-dependencies -print | sed -e 's/\.\///' ); do
```
MichaelSimons: >! -name .git [](start = 61, length = 12)

What do you think about using ".*" instead here?

build-and-test.sh
naamunds: `printf` isn't supported by default on OS X, so I think your proposal is a preferable approach.  I was originally going to do something more like this before realizing that the default find order wasn't sufficient; I didn't think of sorting the dirnames alphabetically, though, so thanks for the suggestion.

---
In reply to: [76320023](https://github.com/dotnet/dotnet-docker-nightly/pull/35#discussion_r76320023) [](ancestors = 76320023)

test/run-test.sh
naamunds: Good idea.

---
In reply to: [76322475](https://github.com/dotnet/dotnet-docker-nightly/pull/35#discussion_r76322475) [](ancestors = 76322475)

netci.groovy
naamunds: Good point.  I've also changed it to use Ubuntu 16.04 machines with the `latest-or-auto-docker` label.

MichaelSimons:
@dotnet-bot test ci please

naamunds:
@dotnet-bot test ci please

naamunds:
@MichaelSimons I've added another commit to address the comments in this PR.  In addition, the commit includes a change to build `*-deps` Dockerfiles first, because I discovered in my testing that the `core` image was incorrectly being built before the `core-deps` image.

build-and-test.sh
```diff
@@ -0,0 +1,24 @@
...
+    for dockerfile_dir in $( egrep -v 'nanoserver|windowsservercore' <<< "${1}" | sort ); do
+        tag="${docker_repo}:$( sed -e 's/.\///' -e 's/debian\///' -e 's/debian/sdk/' -e 's/\//-/g' <<< "${dockerfile_dir}" )"
+        echo "----- Building ${tag} -----"
+        docker build -t "${tag}" "${dockerfile_dir}"
+    done
+}
+
+pushd "${repo_root}" > /dev/null
+
+build_dockerfiles "$( find . -path './.*' -prune -o -name 'Dockerfile' -a -path '*/*-deps/*' -print0 | xargs -0 -n1 dirname )"
```
naamunds: The `xargs -0 -n1 dirname` logic is duplicated here rather than put in the `build_dockerfiles` function because when I put it in `build_dockerfiles`, it seemed that the null characters separating files from `find [...] -print0` were not preserved, causing the `xargs` call to not work correctly.  I didn't find a way to preserve them.

An alternative approach would be to use `-print` instead of `-print0` and then omit `-0` from the `xargs` call, but my understanding is that in that case, spaces and other special characters in file paths may not be handled correctly, so I decided it was preferable to duplicate the calls to `xargs`, especially because it only occurs twice and is a small amount of logic.  I'm open to suggestions on this, though.

MichaelSimons: I wish there was a better way to express the image dependencies.  I guess we can refactor as we go :)

---
In reply to: [76364173](https://github.com/dotnet/dotnet-docker-nightly/pull/35#discussion_r76364173) [](ancestors = 76364173)

PR #49, opened 2016-09-14
Added CI for WindowsServerCore images

MichaelSimons:


MichaelSimons:
test ci please

MichaelSimons:
test ci please

MichaelSimons:
test ci please

MichaelSimons:
cc  @mmitche, @naamunds

mmitche:
LGTM

netci.groovy
```diff
@@ -3,15 +3,26 @@ import jobs.generation.Utilities
...
+def platformList = ['Ubuntu16.04:Debian', 'Windows_2016:WindowsServerCore']
 
-def newJob = job(newJobName) {
-    steps {
-        shell("./build-and-test.sh")
-    }
-}
+platformList.each { platform ->
+	def(hostOS, containerOS) = platform.tokenize(':')
+	def newJobName = Utilities.getFullJobName(project, containerOS, isPr)
```
naamunds: I'd prefer `hostOS`, `containerOS`, and `isPr` to follow the same casing convention.

netci.groovy
```diff
@@ -3,15 +3,26 @@ import jobs.generation.Utilities
...
+
+	def newJob = job(newJobName) {
+		steps {
+			if (hostOS == 'Windows_2016')
+			{
+				shell("powershell -NoProfile -Command .\\build-and-test.ps1")
```
naamunds: Have you verified that the `shell` build step can be used to run a PowerShell script?  The Jenkins help for "Execute shell" build step says that it "Runs a shell script (defaults to sh, but this is configurable) for building the project."

[This article](https://hodgkins.io/automating-with-jenkins-and-powershell-on-windows-part-1) suggests using the PowerShell plugin.  In addition, [this Stack Overflow question](http://stackoverflow.com/questions/10452905/how-to-run-a-powershell-script-as-a-job-in-jenkins) seems to suggest using either the PowerShell plugin or a `batchFile` build step.

The PowerShell plugin is already installed in our Jenkins project; this can be seen by manually adding a "Windows PowerShell" build step at the bottom of a Jenkins job definition (e.g. [this one](http://dotnet-ci.cloudapp.net/job/dotnet_dotnet-docker-nightly/job/master/job/GenPRTest/job/WindowsServerCore_prtest/configure)).  However, I'm not sure how to add a Windows PowerShell build step through a groovy script.

mmitche: Yeah don't use shell for that.  shell actually means *Nix shell.


You can either use batchFile("powershell blah blah") or powershell("blah blah")

build-and-test.ps1
```diff
@@ -0,0 +1,24 @@
...
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference="Stop"
+
+$dockerRepo="microsoft/dotnet-nightly"
+
+pushd $PSScriptRoot
+
+Get-ChildItem -Recurse -Filter Dockerfile | where DirectoryName -Ilike "*\$OS*" | sort DirectoryName | ForEach {
```
naamunds: Consider using the standard name, `Like`, of the `where` parameter, rather than the alias `ILike`, as I think its meaning is more intuitive.  (When reading this code, at first I thought `ILike` was a variant of `Like` with different functionality.)

Also, consider using lower-case for the `foreach` keyword like for the `where` and `sort` keywords.  Functionally-equivalent alternatives are `ForEach-Object` and `%`.

build-and-test.ps1
```diff
@@ -0,0 +1,24 @@
...
+    $tag = "$($dockerRepo):" + $_.DirectoryName.Replace($PSScriptRoot, '').TrimStart('\').Replace('\', '-') -replace "windowsservercore$", "windowsservercore-sdk"
```
naamunds: I'm wondering if it would be better to use the $OS variable rather than hard-coding windowsservercore, because at this point, the "-sdk" tagging convention is the same for all OSes.

`-replace "${OS}$", "${OS}-sdk"`

test/run-test.ps1
```diff
@@ -0,0 +1,65 @@
...
+if ($env:DEBUGTEST -eq $null) {
+    $optionalDockerRunArgs="--rm"
+}
+else {
+    $optionalDockerRunArgs=""
+}
+
+pushd $repoRoot
+
+Get-ChildItem -Directory -Exclude 'test', 'update-dependencies', '.*' | ForEach {
```
naamunds: Same comment as build-and-test.ps1 about `foreach` casing.

test/run-test.ps1
```diff
@@ -0,0 +1,65 @@
...
+    $tagBase="$($dockerRepo):$($_.Name)-$OS"
+
+    $timeStamp = Get-Date -Format o | foreach {$_ -replace ":", "." -replace "+", "-"}
```
naamunds: When I run this locally in PowerShell ISE, I get an exception:

> The regular expression pattern + is not valid.
> At line:1 char:31
> + Get-Date -Format o | foreach {$_ -replace ":", "." -replace "+", "-"}
> +                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : InvalidOperation: (+:String) [], RuntimeException
>     + FullyQualifiedErrorId : InvalidRegularExpression

naamunds: What is the purpose of using `foreach` here?  Would it be clearer to instead do:
`$(Get-Date -Format o) -replace ":", "." -replace "+", "-"`

naamunds: Consider using `Get-Date -Format FileDateTime` instead.  This would be more consistent with the Unix behavior, and would simplify the code, making my previous two comments on this line irrelevant.

test/run-test.ps1
```diff
@@ -0,0 +1,65 @@
...
+Get-ChildItem -Directory -Exclude 'test', 'update-dependencies', '.*' | foreach {
+    $tagBase="$($dockerRepo):$($_.Name)-$OS"
+
+    $timeStamp = Get-Date -Format o | foreach {$_ -replace ":", "." -replace "+", "-"}
+    $appName="app$timeStamp"
+    $appDir="${repoRoot}\.test-assets\${appName}"
+
+    New-Item $appDir -type directory | Out-Null
+
+    Write-Host "----- Testing $tagBase-sdk -----$repoRoot   $appDir"
```
naamunds: Consider adding this extra info ($repoRoot $appDir) to the .sh script as well.

test/run-test.ps1
```diff
@@ -0,0 +1,65 @@
...
+
+    Write-Host "----- Testing $tagBase-core -----"
+    docker run -t $optionalDockerRunArgs -v "$($appDir):c:\$appName" --name "core-test-$appName" --entrypoint dotnet "$tagBase-core" "C:\$appName\publish\$appName.dll"
+    if (-NOT $?) {
+        throw  "Testing $tagBase-core failed"
+    }
+
+    Write-Host "----- Testing $tagBase-onbuild -----"
+    pushd $appDir
+    $onbuildTag = "$tagBase-onbuild".ToLowerInvariant()
```
naamunds: Should this be `"$appName-onbuild"` like in the Bash version?

test/run-test.ps1
```diff
@@ -0,0 +1,65 @@
...
+
+    docker run -t $optionalDockerRunArgs --name "onbuild-test-$appName" $onbuildTag
+    if (-NOT $?) {
+        throw "Testing $tagBase-onbuild failed"
+    }
+
+    if ($env:DEBUGTEST -eq $null) {
+        docker rmi $onbuildTag
+        if (-NOT $?) {
+            exit
```
naamunds: What is the purpose of this?  What would happen if the script didn't explicitly exit on this failure?

netci.groovy
MichaelSimons: Changed isPr to isPR

netci.groovy
MichaelSimons: Changed to use batchFile("powershell ...")

build-and-test.ps1
MichaelSimons: changed to like and made keywords lower-cased

build-and-test.ps1
MichaelSimons: certainly - changed

test/run-test.ps1
MichaelSimons: Fixed

test/run-test.ps1
MichaelSimons: Good suggestion - changed.

test/run-test.ps1
MichaelSimons: I actually removed this - I didn't intend to add this.

test/run-test.ps1
MichaelSimons: The bash version is incorrect IMO.  The purpose of the statement is to convey at a higher level that the xxx-onbuild image is being tested rather than literally stating the image that is being used to test with.  I will update the Bash version.

test/run-test.ps1
MichaelSimons: Changed exit to throw with a message indicating the source of the failure.

test/run-test.ps1
MichaelSimons: yes - changed.

MichaelSimons:
test ci please

naamunds:
:shipit:

PR #51, opened 2016-09-20
Enable CI on Nano Server

MichaelSimons:
cc @naamunds

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -8,21 +8,17 @@ def platformList = ['Ubuntu16.04:Debian', 'Windows_2016:WindowsServerCore', 'Win
 platformList.each { platform ->
     def(hostOS, containerOS) = platform.tokenize(':')
     def newJobName = Utilities.getFullJobName(project, containerOS, isPR)
-    def machineLabel = 'latest-or-auto-docker'
+    def machineLabel = (hostOS == 'Windows Nano 2016') ? '' : 'latest-or-auto-docker'
 
     def newJob = job(newJobName) {
         steps {
-            if (hostOS == 'Windows_2016')
-            {
-                batchFile("powershell -NoProfile -Command .\\build-and-test.ps1")
+            if (hostOS == 'Windows_2016') {
+                batchFile("powershell -NoProfile -Command .\\build-and-test.ps1 -OS windowsservercore")
             }
-            else if (hostOS == 'Windows Nano 2016')
-            {
+            else if (hostOS == 'Windows Nano 2016') {
                 batchFile("powershell -NoProfile -Command .\\build-and-test.ps1 -OS nanoserver")
-                machineLabel = ''
             }
-            else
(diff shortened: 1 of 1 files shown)


MichaelSimons:
test ci please

MichaelSimons:
test ci please

netci.groovy
```diff
@@ -3,26 +3,32 @@ import jobs.generation.Utilities
...
+    def newJob = job(newJobName) {
+        steps {
+            if (hostOS == 'Windows_2016')
+            {
+                batchFile("powershell -NoProfile -Command .\\build-and-test.ps1")
+            }
+            else if (hostOS == 'Windows Nano 2016')
+            {
+                batchFile("powershell -NoProfile -Command .\\build-and-test.ps1 -OS nanoserver")
+                machineLabel = ''
```
naamunds: It feels a bit strange to me to set `machineLabel` inside the definition of the job's steps.  What do you think about just initializing it to its correct value by using the ternary operator on line 11?

netci.groovy
```diff
@@ -3,26 +3,32 @@ import jobs.generation.Utilities
...
-			{
-				shell("./build-and-test.sh")
-			}
-		}
-	}
+    def newJob = job(newJobName) {
+        steps {
+            if (hostOS == 'Windows_2016')
+            {
+                batchFile("powershell -NoProfile -Command .\\build-and-test.ps1")
```
naamunds: Would it preferable to use the `OS` parameter here also, rather than relying on its default value?

netci.groovy
```diff
@@ -3,26 +3,32 @@ import jobs.generation.Utilities
...
-			else
-			{
-				shell("./build-and-test.sh")
-			}
-		}
-	}
+    def newJob = job(newJobName) {
+        steps {
+            if (hostOS == 'Windows_2016')
+            {
```
naamunds: Consider putting this brace on the previous line for formatting consistency.  Same with the ones for "else if" and "else".

MichaelSimons: Fixed

---
In reply to: [79723933](https://github.com/dotnet/dotnet-docker-nightly/pull/51#discussion_r79723933) [](ancestors = 79723933)

netci.groovy
MichaelSimons: Agreed - fixed

---
In reply to: [79722462](https://github.com/dotnet/dotnet-docker-nightly/pull/51#discussion_r79722462) [](ancestors = 79722462)

netci.groovy
MichaelSimons: Not a big deal but changed as suggested.

---
In reply to: [79722600](https://github.com/dotnet/dotnet-docker-nightly/pull/51#discussion_r79722600) [](ancestors = 79722600)

MichaelSimons:
test ci please

PR #58, opened 2016-09-27
Updated Windows Dockerfiles to build 10.0.14393.206

MichaelSimons:
With this version the forwarders workaround is no longer needed.

cc @naamunds


naamunds:
:shipit:

PR #62, opened 2016-09-28
Modified Nano Dockerfiles to make use of new Invoke-WebRequest support

MichaelSimons:
@naamunds


naamunds:
It's great that the nanoserver and windowsservercore Dockerfiles are now identical except for the `FROM` line.

:shipit:

PR #64, opened 2016-09-30
Add ca-certificates to core-deps

naamunds:
See https://github.com/dotnet/dotnet-docker/issues/121.

cc: @MichaelSimons @glennc

### DIFF
--- a/rel-1.0.0/debian/core-deps/Dockerfile
+++ b/rel-1.0.0/debian/core-deps/Dockerfile
@@ -3,6 +3,7 @@ FROM debian:jessie
 # Install .NET Core dependencies
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
+        ca-certificates \
         libc6 \
         libcurl3 \
         libgcc1 \
--- a/rel-1.0.0/debian/core-deps/Dockerfile
+++ b/rel-1.0.0/debian/core-deps/Dockerfile
@@ -1,9 +1,10 @@
 FROM debian:jessie
 
-# Install .NET Core dependencies
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         ca-certificates \
+        \
+# .NET Core dependencies
         libc6 \
         libcurl3 \
         libgcc1 \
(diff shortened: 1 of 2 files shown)


rel-1.0.0/debian/core/Dockerfile
```diff
@@ -1,9 +1,7 @@
 FROM microsoft/dotnet-nightly:rel-1.0.0-core-deps
 
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-        ca-certificates \
-        curl \
```
MichaelSimons: >curl \ [](start = 8, length = 6)

My preference would be to leave `curl \` on a separate line.  It makes it visually easier to read the packages that are installed and makes adding/removing packages easier to diff.

rel-1.0.0/debian/core-deps/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM debian:jessie
 # Install .NET Core dependencies
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
+        ca-certificates \
```
MichaelSimons: The comment above this section with this new dependency is no correct.  This is not a .NET Core dependency.  I have seen this type of situation (wanting to call out/group what the various dependencies are for) handled in a variety of ways within official Dockerfiles

1. Use separate layers for different kinds of dependencies
2. User "variables" to define the different kinds of dependencies and then one apt-get update statement that references the variables.
3.  Use one apt-get update statement with comments separating out the different kind of dependencies.

Each one has some pros and cons.  I personally prefer #3 for this Dockerfile.

RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        ca-certificates \
        \
# .NET Core dependencies
        libc6 \
        libcurl3 \
        libgcc1 \
        libgssapi-krb5-2 \
        libicu52 \
        liblttng-ust0 \
        libssl1.0.0 \
        libstdc++6 \
        libunwind8 \
        libuuid1 \
        zlib1g \

rel-1.0.0/debian/core/Dockerfile
naamunds: Updated as suggested.

rel-1.0.0/debian/core-deps/Dockerfile
naamunds: I agree that your proposal makes the most sense for this Dockerfile. I've updated it to match that.

PR #76, opened 2016-10-12
Added Dockerfiles for rel-1.0.0-preview2.1 branch of CLI

MichaelSimons:
cc @naamunds

### DIFF
--- a/rel-1.0.0-preview2.1/debian/core-deps/hooks/post_push
+++ b/rel-1.0.0-preview2.1/debian/core-deps/hooks/post_push
@@ -11,7 +11,7 @@ if [[ $DOCKERCFG == {\"auths\":* ]]; then
 fi
 
 
-echo "Pushing core images"
+echo "Pushing core image"
 versionedImageName=$repoName":rel-1.0.0-preview2.1-core"
 docker tag core $versionedImageName
 docker push $versionedImageName
--- a/rel-1.0.0-preview2.1/debian/core-deps/hooks/post_push
+++ b/rel-1.0.0-preview2.1/debian/core-deps/hooks/post_push
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+# Issue - https://github.com/docker/hub-feedback/issues/384#issuecomment-214508878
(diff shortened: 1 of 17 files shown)


rel-1.0.0-preview2.1/debian/core-deps/hooks/post_push
```diff
@@ -0,0 +1,17 @@
...
+# Issue - https://github.com/docker/hub-feedback/issues/384#issuecomment-214508878
+# Docker Hub emits an invalid .dockercfg.  Remove the outer auths element.
+if [[ $DOCKERCFG == {\"auths\":* ]]; then
+    jsonStartIndex=$(expr index "$DOCKERCFG" :)
+    length=$((${#DOCKERCFG} - $jsonStartIndex - 1))
+    echo ${DOCKERCFG:jsonStartIndex:length} > /root/.dockercfg
+fi
+
+
+echo "Pushing core images"
```
naamunds: This should be "image" instead of "images". Same for the onbuild image in rel-1.0.0-preview2.1/debian/hooks/post_build.

test/create-run-publish-app.sh
```diff
@@ -12,6 +13,10 @@ dotnet run
 dotnet publish -o publish/framework-dependent
 
 echo "Testing self-contained deployment"
+if [ "rel-1.0.0-preview2.1" == "${2}" ]; then
```
naamunds: It seems like the logic for checking the image version fits better in run-test.sh.  What do you think about passing the path to the NuGet.Config as an optional second argument to create-run-publish.sh, rather than passing the sdk image version?  run-test.sh would pass in the path if and only if `development_image_version` equals `rel-1.0.0-preview2.1`.

rel-1.0.0-preview2.1/debian/core-deps/hooks/post_push
MichaelSimons: Fixed

test/create-run-publish-app.sh
MichaelSimons: This approach would add more complexity.  Leaving as is for now.

PR #77, opened 2016-10-13
Fixed broken build hooks

MichaelSimons:


MichaelSimons:
Merging as this is needed right away.

PR #81, opened 2016-10-20
Fix build and test script to work when parent dir contains 'nanoserver'

MichaelSimons:


MichaelSimons:
cc @naamunds

naamunds:
:shipit:

PR #82, opened 2016-10-22
Update onbuild - remove 'run'

richlander:
In response to https://github.com/dotnet/dotnet-docker/pull/134/files

Tested the debian images this time. Not on a Windows machine right now, so hoping it acts the same.

/cc @MichaelSimons, @kendrahavens

### DIFF
--- a/rel-1.0.0-preview2.1/debian/onbuild/Dockerfile
+++ b/rel-1.0.0-preview2.1/debian/onbuild/Dockerfile
@@ -1,9 +1,10 @@
 FROM microsoft/dotnet-nightly:rel-1.0.0-preview2.1-sdk
 
-RUN mkdir -p /dotnetapp
 WORKDIR /dotnetapp
 
-ENTRYPOINT ["dotnet", "run"]
+ENTRYPOINT ["dotnet", "out/dotnetapp.dll"]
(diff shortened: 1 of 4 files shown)


rel-1.0.0-preview2.1/debian/onbuild/Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM microsoft/dotnet-nightly:rel-1.0.0-preview2.1-sdk
 
-RUN mkdir -p /dotnetapp
 WORKDIR /dotnetapp
 
-ENTRYPOINT ["dotnet", "run"]
+ENTRYPOINT ["dotnet", "out/dotnetapp.dll"]
```
MichaelSimons: Regarding my previous concern about hard coding the assembly name here.  This is also going to break as soon as we take a csproj based CLI.  With csproj's, the default assembly name is the name of the csproj, not the containing folder.  Additionally, you can specify the `AssemblyName` property to change this behavior.  I am not seeing a way for a simple entry point, we are going to need logic in order to determine the assembly name.

MichaelSimons:
onbuild images are going to be discontinued - closing

PR #94, opened 2016-10-31
Update Debian sdk Dockerfile to CLI 1.0.0-preview3-003933

MichaelSimons:
@naamunds


MichaelSimons:
CI failure is expected because of the "3" verbs becoming the default verbs.  CI scripts need to be updated.

MichaelSimons:
Merging to unblock aspnet - will fix CI shortly.

PR #95, opened 2016-10-31
CI changes needed for latest CLI where csproj is the default experience

MichaelSimons:


test/create-run-publish-app.sh
```diff
@@ -33,5 +25,8 @@ if [ "rel-1.0.0-preview2.1" == "${2}" ]; then
     dotnet run
     dotnet publish -o publish/self-contained
 else
-    dotnet publish3 -r debian.8-x64 -o publish/self-contained
+    sed -i '/<PropertyGroup>/a \    <RuntimeIdentifiers>debian.8-x64<\/RuntimeIdentifiers>' ./${1}.csproj
```
naamunds: >./${1}.csproj [](start = 92, length = 13)

It looks like this assumes that the directory passed as $1 is at the root level (e.g. "/app123" or "app123").  If the directory is not at the root level, like "/test/app123", then I assume this code wouldn't work.

This is fine with the current usage of this script by run-test.sh, but consider making it more robust.

naamunds: At minimum, adding a comment mentioning that $1 should be at the root level would be helpful.

naamunds: `>/a \`
What is the purpose of the backslash here?  Would the spaces following it otherwise not be written out?

PR #98, opened 2016-11-01
Misc update-dependencies fixes.

MichaelSimons:
1. Modified update-dependencies to use a specific version of CLI.  Previously it was using the latest build which broke when CLI moved to msbuild.
2. Migrated the update-dependencies app to be csproj based.
3. Modified the update-dependencies app to be portable (simplification).
4. Modified update-dependencies to use the FileRegexReleaseUpdater to be more flexible since the packages have changed between CLI channels.

@dotnet-bot skip CI please

@naamunds


naamunds:
:shipit:

PR #103, opened 2016-11-03
Updated update-dependencies to work with latest Dockerfile refactoring

MichaelSimons:
This requires an update to the Maestro subscriptions as well.

@dotnet-bot skip CI please

@naamunds

### DIFF
--- a/update-dependencies/Program.cs
+++ b/update-dependencies/Program.cs
@@ -108,7 +108,7 @@ namespace Dotnet.Docker.Nightly
             {
                 Path = path,
                 BuildInfoName = "Cli",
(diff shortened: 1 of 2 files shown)


update-dependencies/Program.cs
```diff
@@ -109,7 +108,7 @@ private static IDependencyUpdater CreateDependencyUpdater(string path)
             {
                 Path = path,
                 BuildInfoName = "Cli",
-                Regex = new Regex($@"ENV DOTNET_SDK_VERSION (?<version>[^\r\n]*)"),
+                Regex = new Regex($@"ENV DOTNET_SDK_VERSION [\d.]*-(?<version>{s_config.CliReleaseMoniker}-[^\r\n]*)"),
```
naamunds: Should this be `\d\.` to only match digits and the '.' character rather than matching all characters with the unescaped `.`?

PR #108, opened 2016-11-04
Update Debian sdk Dockerfiles to CLI 1.0.0-preview3-004050

naamunds:
/cc @dleeapho @MichaelSimons


dleeapho:
LGTM, once the Debian tests pass that is good (NanoServer is a known issue).

PR #110, opened 2016-11-07
Fixed incorrect tag listed for 1.0.1-sdk-projectjson-nanoserver

MichaelSimons:
@dotnet-bot skip CI please

@naamunds


MichaelSimons:
Rebased changes with master wince there was a merge conflict.

naamunds:
In addition to the change you have here, I found another discrepancy with the tags.  The `1.0-sdk` tag is documented in the README but does not exist in Docker Hub; DH has `1.0-sdk-msbuild` instead.  However, `1.0-sdk-nanoserver` is in both the README and Docker Hub.  This seems inconsistent; I think the tags should either be `1.0-sdk-msbuild` and `1.0-sdk-msbuild-nanoserver`, or `1.0-sdk` and `1.0-sdk-nanoserver`.

This comments applies to the 1.1 versions of these tags in addition to the 1.0 versions.

MichaelSimons:
@naamunds - `1.0-sdk` and `1.0-sdk-nanoserver` are the tags we should have.  I will fix the builds settings accordingly.

naamunds:
@MichaelSimons  Sounds good.

It looks like the commit and PR title still say "1.0.2".  Other than that, LGTM.

PR #125, opened 2016-11-18
Update rel-1.0.0-preview2 SDK to preview2-003154

MichaelSimons:


MichaelSimons:
Testing update-dependencies

PR #126, opened 2016-11-18
Modified update-dependencies logic to handle distinguishing preview2 from preview2-1

MichaelSimons:
I discovered this issue while adding a Maestro hook to update preview2.  It was also replacing the preview2-1 version which is incorrect.

@dotnet-bot skip CI please

@naamunds

### DIFF
--- a/update-dependencies/Program.cs
+++ b/update-dependencies/Program.cs
@@ -108,7 +108,7 @@ namespace Dotnet.Docker.Nightly
             {
                 Path = path,
                 BuildInfoName = "Cli",
-                Regex = new Regex($@"ENV DOTNET_SDK_VERSION [\d\.]*-(?<version>{s_config.CliReleaseMoniker}-[\d]*)\r\n"),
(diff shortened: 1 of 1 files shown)


update-dependencies/Program.cs
```diff
@@ -108,7 +108,7 @@ private static IDependencyUpdater CreateDependencyUpdater(string path)
             {
                 Path = path,
                 BuildInfoName = "Cli",
-                Regex = new Regex($@"ENV DOTNET_SDK_VERSION [\d\.]*-(?<version>{s_config.CliReleaseMoniker}-[^\r\n]*)"),
+                Regex = new Regex($@"ENV DOTNET_SDK_VERSION [\d\.]*-(?<version>{s_config.CliReleaseMoniker}-[\d]*)\r\n"),
```
naamunds: Could we omit the brackets from `[\d]*` and just use `\d*`?

naamunds: Consider using the '?' quantifier for '\r', in case the file just has LF instead of CRLF:  `\r?\n`

MichaelSimons: won't fix

MichaelSimons: After thinking about this I changed to \d+

PR #130, opened 2016-11-21
Introduce 1.0.3 artifacts

MichaelSimons:
@naamunds

### DIFF
--- a/1.0/debian/sdk/msbuild/Dockerfile
+++ b/1.0/debian/sdk/msbuild/Dockerfile
@@ -16,21 +16,34 @@ RUN apt-get update \
         zlib1g \
     && rm -rf /var/lib/apt/lists/*
 
-# Install .NET Core SDK
-ENV DOTNET_SDK_VERSION 1.0.0-preview3-004056
-ENV DOTNET_SDK_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DOTNET_SDK_VERSION/dotnet-dev-debian-x64.$DOTNET_SDK_VERSION.tar.gz
+# Install .NET Core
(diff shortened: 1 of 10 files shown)


MichaelSimons:
@dotnet-bot test NanoServer please

1.0/debian/sdk/msbuild/Dockerfile
```diff
@@ -16,21 +16,34 @@ RUN apt-get update \
...
+# Install .NET Core
+ENV DOTNET_VERSION 1.0.3
+ENV DOTNET_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/$DOTNET_VERSION/dotnet-debian-x64.$DOTNET_VERSION.tar.gz
+
+RUN curl -SL $DOTNET_DOWNLOAD_URL --output dotnet.tar.gz \
+    && mkdir -p /usr/share/dotnet \
+    && tar -zxf dotnet.tar.gz -C /usr/share/dotnet \
+    && rm dotnet.tar.gz \
+    && ln -s /usr/share/dotnet/dotnet /usr/bin/dotnet
+
```
naamunds: Is there going to be an MSBuild SDK version that has 1.0.3 so that we won't have to explicitly install it and have the manual cache population logic? I think we should definitely have this logic as you have it here if it's not available yet, but I'm wondering if it will be available by the time we introduce these artifacts in the official repo.

MichaelSimons: @naamunds, No there won't be an msbuild version that targets 1.0.3.

MichaelSimons:
@naamunds, good catch on the hook - updated.

PR #131, opened 2016-11-28
Re-order versions

vivmishra:
Reordering the images in the ReadMe so that they appear ordered as per version first rather than as per product. This makes it easier to spot all the desired images for a version.

@dotnetbot skip CI please


MichaelSimons:
@vivmishra, We should capture the reason for this change someplace.  I would suggest including it in the PR description of via a linked issue.  Also for PRs like this one that do not impact artifacts that get tested by CI, you should include a `@dotnetbot skip CI please` comment in the description.  This will indicate CI should not be run on this PR.  It is a little optimization that can save the amount of resources used by CI.

MichaelSimons:
@richlander - Please take a look to ensure this is the sort order you were expecting.

PR #132, opened 2016-11-28
Modified build-and-test.ps1 to be xplat

MichaelSimons:
This is the first iteration towards having one set of CI scripts for all supported platforms.  The next step would be to get Jenkins to install Powershell and utilize these changes.  Once this is done the bash scripts can be removed and we then only have one set of scripts to maintain.

@naamunds

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -3,7 +3,7 @@ param(
     [Parameter(Mandatory=$true)]
     [ValidateSet("win", "linux")]
     [string]$Platform,
-    [switch]$UseImageCache=$false
+    [switch]$UseImageCache
 )
 
 Set-StrictMode -Version Latest
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -1,19 +1,57 @@
+[cmdletbinding()]
+param(
+    [Parameter(Mandatory=$true)]
+    [ValidateSet("win", "linux")]
+    [string]$Platform,
+    [switch]$UseImageCache=$false
+)
+
 Set-StrictMode -Version Latest
 $ErrorActionPreference="Stop"
 
 $dockerRepo="microsoft/dotnet-nightly"
+$dirSeparator = [IO.Path]::DirectorySeparatorChar
+
+if ($UseImageCache) {
+    $optionalDockerBuildArgs=""
+}
+else {
+    $optionalDockerBuildArgs = "--no-cache"
+}
+
+if ($Platform -eq "win") {
+    $imageOs = "nanoserver"
+    $tagSuffix = "-nanoserver"
+}
+else {
+    $imageOs = "debian"
+    $tagSuffix = ""
+}
 
 pushd $PSScriptRoot
 
-Get-ChildItem -Recurse -Filter Dockerfile | where {$_.DirectoryName.TrimStart($PSScriptRoot) -like "*\nanoserver*"} | sort DirectoryName | foreach {
-    $tag = "$($dockerRepo):" + $_.DirectoryName.Replace($PSScriptRoot, '').Replace("\nanoserver", '').TrimStart('\').Replace('\', '-') + "-nanoserver"
-    Write-Host "--- Building $tag from $($_.DirectoryName) ---"
-    docker build --no-cache -t $tag $_.DirectoryName
-    if (-NOT $?) {
-        throw "Failed building $tag"
+$tags = [System.Collections.ArrayList]@()
+Get-ChildItem -Recurse -Filter Dockerfile |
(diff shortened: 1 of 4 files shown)


MichaelSimons:
@dotnet-bot test ci please

MichaelSimons:
@dotnet-bot test NanoServer please

test/run-test.ps1
```diff
@@ -1,42 +1,89 @@
...
-    New-Item $appDir -type directory | Out-Null
+        New-Item $appDir -type directory | Out-Null
 
-    Write-Host "----- Testing $fullSdkTag -----"
-    docker run -t $optionalDockerRunArgs -v "$($appDir):c:\$appName" -v "$repoRoot\test:c:\test" --name "sdk-test-$appName" --entrypoint powershell "$fullSdkTag" c:\test\create-run-publish-app.ps1 "c:\$appName" "${sdkTag}"
-    if (-NOT $?) {
-        throw  "Testing $full_sdk_tag failed"
-    }
+        Write-Host "----- Testing $fullSdkTag -----"
+        docker run -it --rm `
```
naamunds: Why does this now use the `-i` flag?

test/run-test.ps1
```diff
@@ -1,42 +1,89 @@
...
+            -v "${appDir}:${containerRoot}${appName}" `
+            -v "${repoRoot}${dirSeparator}test:${containerRoot}test" `
+            --name "sdk-test-$appName" `
+            $fullSdkTag $sdkRunArg "${containerRoot}test${dirSeparator}create-run-publish-app.${testScriptSuffix}" "${containerRoot}${appName}" $sdkTag
```
naamunds: Is `--entrypoint` no longer necessary here?

test/run-test.ps1
```diff
@@ -1,42 +1,89 @@
...
+            throw  "Testing $baseTag-runtime failed"
+        }
+
+        if ($Platform -eq "linux") {
+            Write-Host "----- Testing $baseTag-runtime-deps with $sdkTag app -----"
+            docker run -t --rm `
+                -v "${appDir}:${containerRoot}${appName}" `
+                --name "runtime-deps-test-$appName" `
+                --entrypoint "${containerRoot}${appName}${dirSeparator}publish${dirSeparator}self-contained${dirSeparator}${appName}" `
+                "$baseTag-runtime-deps"
```
naamunds: What do you think about adding `$tagSuffix` here and in the `Write-Host` and `throw` statements?  This isn't necessary today, but if we add other platforms that have a runtime-deps image such as Alpine, it would be necessary, and I don't really see a significant downside to adding it.

test/run-test.ps1
```diff
@@ -1,42 +1,89 @@
...
-        throw  "Testing $baseTag-runtime failed"
+        Write-Host "----- Testing $baseTag-runtime$tagSuffix with $sdkTag app -----"
+        docker run -t --rm `
+            -v "${appDir}:${containerRoot}${appName}" `
+            --name "runtime-test-$appName" `
+            --entrypoint dotnet `
+            "$baseTag-runtime$tagSuffix" `
+            "${containerRoot}${appName}${dirSeparator}publish${dirSeparator}framework-dependent${dirSeparator}${appName}.dll"
+        if (-NOT $?) {
+            throw  "Testing $baseTag-runtime failed"
```
naamunds: This should be `$baseTag-runtime$tagSuffix`.

build-and-test.ps1
```diff
@@ -1,19 +1,57 @@
+[cmdletbinding()]
+param(
+    [Parameter(Mandatory=$true)]
+    [ValidateSet("win", "linux")]
+    [string]$Platform,
```
MichaelSimons: I would prefer to not have a default because using the default in a linux environment wastes a lot of time because it triggers the nano images to be pulled which fails but not immediately.

build-and-test.ps1
```diff
@@ -1,19 +1,57 @@
...
+    [switch]$UseImageCache=$false
```
MichaelSimons: Removed initialization.

test/run-test.ps1
MichaelSimons: It was moved to the $sdkRunArg.  This made it easier to handle because it is not needed for linux.

test/run-test.ps1
MichaelSimons: fixed

test/run-test.ps1
MichaelSimons: Removed - a leftover from debugging something.

PR #133, opened 2016-11-29
Add builds hooks to produce sdk multi-tags

### LINKED ISSUE #112: Some images built from the same Dockerfile have different layers
This repo has several images which have multiple tags; for example, `1.0.1-runtime` and `1.0-runtime` point to the same image.  For some images, we are assigning the same image multiple tags, either by the build hooks or the Nano Server VSTS build definitions.  For other images, we have multiple entries in the Docker Hub Build Settings for the same Dockerfiles to be built with different tags.

In the former case, the image digests are the same (meaning the tags point to the exact same image). In the latter case, the expectation is that the images should have different digests but contain the same layers.  However, this is currently not the case for some images; for example, `microsoft/dotnet-nightly:1.0-sdk-msbuild` and `microsoft/dotnet-nightly:1.0.1-sdk-msbuild` contain three different layers, which can be seen with the `docker inspect` command.  This means that users who pull both tags will have separate layers for them on their machines, which takes up a substantial amount of extra space and increases download times.

MichaelSimons:
fixes #112 

@naamunds

### DIFF
--- a/1.0/debian/runtime-deps/hooks/post_push
+++ b/1.0/debian/runtime-deps/hooks/post_push
@@ -2,47 +2,35 @@
 
 set -e
 
-# Issue - https://github.com/docker/hub-feedback/issues/384#issuecomment-214508878
-# Docker Hub emits an invalid .dockercfg.  Remove the outer auths element.
-if [[ $DOCKERCFG == {\"auths\":* ]]; then
-    jsonStartIndex=$(expr index "$DOCKERCFG" :)
-    length=$((${#DOCKERCFG} - $jsonStartIndex - 1))
-    echo ${DOCKERCFG:jsonStartIndex:length} > /root/.dockercfg
-fi
-
-
-echo "Pushing runtime-deps images"
-tagStartIndex=$(expr index "$IMAGE_NAME" :)
-repoName=${IMAGE_NAME:0:tagStartIndex - 1}
-
-imageName=$repoName":1.0.3-runtime-deps"
+echo "Pushing additional runtime-deps tags"
+imageName=$DOCKER_REPO":1.0.3-runtime-deps"
 docker tag $IMAGE_NAME $imageName
 docker push $imageName
 
 # Existing core-deps tags are obsolete but being preserved for through the 1* release
-imageName=$repoName":1.0.3-core-deps"
+imageName=$DOCKER_REPO":1.0.3-core-deps"
 docker tag $IMAGE_NAME $imageName
 docker push $imageName
 
-imageName=$repoName":1.0-core-deps"
+imageName=$DOCKER_REPO":1.0-core-deps"
 docker tag $IMAGE_NAME $imageName
 docker push $imageName
 
 
-echo "Pushing runtime images"
-imageName=$repoName":1.0.3-runtime"
+echo "Pushing additional runtime tags"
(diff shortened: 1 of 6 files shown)


1.0/debian/sdk/msbuild/hooks/post_push
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+echo "Pushing additional sdk-msbuild tags"
```
MichaelSimons: @naamunds, I am guessing you are going to call out the usage of `tags` in the comment when only one additional tag is pushed.  I am proposing plural as the number of tags we have for images are constantly changing.  Changing this comment constantly between singular and plural adds such little value IMO.  I think the comment as is conveys the intent of the code unless you have a wording that would be 100% correct and doesn't need updating between singular and plural images.

1.0/debian/runtime-deps/hooks/post_push
```diff
@@ -2,47 +2,35 @@
...
 
-imageName=$repoName":1.0-core-deps"
+imageName=$DOCKER_REPO":1.0-core-deps"
 docker tag $IMAGE_NAME $imageName
 docker push $imageName
 
 
-echo "Pushing runtime images"
-imageName=$repoName":1.0.3-runtime"
+echo "Pushing additional runtime tags"
```
naamunds: I think I'd prefer this to say "Pushing runtime image" because without this hook, no runtime images are on Docker Hub.  "Pushing additional runtime tags" implies to me that the runtime image is already there, which is the case for runtime-deps but not runtime.  This applies to the 1.1 post_push file also.

1.0/debian/sdk/msbuild/hooks/post_push
naamunds: I'm fine with this wording.  Could use "tag(s)", but "tags" is fine too.

1.1/debian/sdk/projectjson/hooks/post_push
```diff
@@ -0,0 +1,16 @@
...
+docker tag $IMAGE_NAME $imageName
+docker push $imageName
+
+imageName=$DOCKER_REPO":sdk"
+docker tag $IMAGE_NAME $imageName
+docker push $imageName
+
+imageName=$DOCKER_REPO":latest"
+docker tag $IMAGE_NAME $imageName
+docker push $imageName
```
naamunds: Currently the 1.0/debian/sdk/projectjson Dockerfile is tagged as latest in dotnet-nightly.  However, the README says that 1.1 is latest, and the dotnet repo uses 1.1 as latest, so I'm assuming that's a bug that we hadn't caught, which this will fix.

MichaelSimons: Correct.

1.0/debian/runtime-deps/hooks/post_push
MichaelSimons: Fixed as suggested.

1.0/debian/sdk/msbuild/hooks/post_push
MichaelSimons: I had thought about that.  I am going to stick with tags.

PR #147, opened 2017-01-06
Upgrading to CLI 1.0.0-preview5-004432

MichaelSimons:
CLI is now based on 1.0.3 therefore we no longer need to explicitly install the 1.0.3 runtime for the 1.0 images.  Additionally the latest CLI now supports Nano Server again therefore the nano images were updated to use the latest preview5 CLI.

### DIFF
--- a/1.1/nanoserver/sdk/msbuild/Dockerfile
+++ b/1.1/nanoserver/sdk/msbuild/Dockerfile
@@ -29,7 +29,7 @@ RUN mkdir warmup \
     && cd warmup \
     && dotnet new \
     # Projects created with .NET Core SDK preview5 target 1.0, replace with 1.1 to populate cache
-    && powershell -NoProfile -Command "(Get-Content warmup.csproj).replace('1.0.1', '1.1.0').replace('netcoreapp1.0', 'netcoreapp1.1') | Set-Content warmup.csproj" \
+    && powershell -NoProfile -Command "(Get-Content warmup.csproj).replace('1.0.3', '1.1.0').replace('netcoreapp1.0', 'netcoreapp1.1') | Set-Content warmup.csproj" \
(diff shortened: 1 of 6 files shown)


1.1/nanoserver/sdk/msbuild/Dockerfile
```diff
@@ -28,8 +28,8 @@ ENV DOTNET_SKIP_FIRST_TIME_EXPERIENCE true
 RUN mkdir warmup \
     && cd warmup \
     && dotnet new \
-    # .NET Core SDK preview3 targets 1.0, replace with 1.1 to populate cache
-    && powershell -NoProfile -Command "(Get-Content project.json).replace('1.0.1', '1.1.0').replace('netcoreapp1.0', 'netcoreapp1.1') | Set-Content project.json" \
+    # Projects created with .NET Core SDK preview5 target 1.0, replace with 1.1 to populate cache
+    && powershell -NoProfile -Command "(Get-Content warmup.csproj).replace('1.0.1', '1.1.0').replace('netcoreapp1.0', 'netcoreapp1.1') | Set-Content warmup.csproj" \
```
naamunds: Should this be replacing 1.0.3 instead of 1.0.1?  When running `dotnet new` with 1.0.0-preview5-004432, the csproj contains the following line:
```
<PackageReference Include="Microsoft.NETCore.App" Version="1.0.3" />
```

naamunds: This comment applies to the debian Dockerfile as well, and also to the create-run-publish-app scripts.

MichaelSimons: Good catch - fixed

naamunds:
:shipit:

PR #152, opened 2017-01-11
Removed explicit install of 1.1 runtime in 1.1-sdk-msbuild Dockerfiles

MichaelSimons:
This is no longer necessary because of https://github.com/dotnet/cli/pull/5225


rakeshsinghranchi:
Looks good to me.

PR #153, opened 2017-01-11
Refactored tags for the RC3 release of CLI

MichaelSimons:


MichaelSimons:
@rakeshsinghranchi, yes the build def has been updated accordingly.  Thanks

PR #155, opened 2017-01-13
Utilize dotnet package warmup logic for rc3

MichaelSimons:
With the latest changes CLI made to install both the current and LTS runtime, the warmup logic was also modified to initialize the NuGet package cache for both.  Because of this, the logic needed to trigger the cache warmup for current is not needed therefore it was removed.  I also tweaked the test code to make use of the new -t console1.1 logic for simplification.

### DIFF
--- a/1.1/debian/sdk/msbuild/Dockerfile
+++ b/1.1/debian/sdk/msbuild/Dockerfile
@@ -28,13 +28,9 @@ RUN curl -SL $DOTNET_SDK_DOWNLOAD_URL --output dotnet.tar.gz \
 
 # Trigger the population of the local package cache
 ENV NUGET_XMLDOC_MODE skip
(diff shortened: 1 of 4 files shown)


naamunds:
How much disk space do the package caches take?  I'm wondering if we should consider continuing to only populate the 1.1 cache for the 1.1 Docker images if populating both 1.0 and 1.1 would bloat the image.  In that case, we should also consider using `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` to only populate the 1.0 cache for the 1.0 images.

MichaelSimons:
@naamunds, A decision was made to not remove the unused shared framework and to use the standard CLI behavior for RC3.  CLI is working getting aligned with the runtime therefore there will only be one version of the shared framework installed.

PR #156, opened 2017-01-13
Update CLI to 1.0.0-rc3-004500

MichaelSimons:
Updating manually since CLI is not updating their [versions file](https://github.com/dotnet/versions/tree/master/build-info/dotnet/cli/rel/1.0.0-rc3) therefore Maestro is not doing this.


naamunds:
:shipit:

PR #171, opened 2017-02-03
Updating to CLI to 1.0.0-rc4-004756

MichaelSimons:
The latest CLI introduced [new `dotnet new` functionality](2d93968a88a724ec69a3c3cfd0ad92576101cbfe) therefore the tests needed to be updated accordingly.

### DIFF
--- a/test/run-test.ps1
+++ b/test/run-test.ps1
@@ -63,7 +63,10 @@ Get-ChildItem -Recurse -Filter Dockerfile |
         $buildImage = "sdk-build-$appName"
         $dotnetNewParam = ""
         if ($sdkTag -like "*1.1-sdk-msbuild*") {
-            $dotnetNewParam = "-t Console1.1"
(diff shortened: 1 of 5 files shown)


test/run-test.ps1
```diff
@@ -63,7 +63,10 @@ Get-ChildItem -Recurse -Filter Dockerfile |
         $buildImage = "sdk-build-$appName"
         $dotnetNewParam = ""
         if ($sdkTag -like "*1.1-sdk-msbuild*") {
-            $dotnetNewParam = "-t Console1.1"
+            $dotnetNewParam = "console --Framework netcoreapp1.1"
+        }
+        elseif ($sdkTag -like "*1.0-sdk-msbuild*") {
+            $dotnetNewParam = "console --Framework netcoreapp1.0"
```
naamunds: Consider factoring out the commonality between these two values of `dotnetNewParam`, as the only difference now is "1.1" vs. "1.0" at the end.

PR #182, opened 2017-02-07
Fixup tests to run against 2.0 assets

MichaelSimons:


test/run-test.ps1
```diff
@@ -42,7 +42,7 @@ pushd $repoRoot
 
 # Loop through each sdk Dockerfile in the repo and test the sdk and runtime images.
 Get-ChildItem -Recurse -Filter Dockerfile |
-    where DirectoryName -like "*${dirSeparator}${imageOs}${dirSeparator}sdk${dirSeparator}*" |
+    where DirectoryName -like "*${dirSeparator}${imageOs}${dirSeparator}sdk*" |
```
naamunds: What prompted this change?

test/run-test.ps1
```diff
@@ -131,6 +131,15 @@ Get-ChildItem -Recurse -Filter Dockerfile |
                     }
                 }
 
+                if ($sdkTag -like "*2.0-sdk") {
+                    # Temporary workaround https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/dogfooding.md#option-2-self-contained
+                    exec { docker run --rm `
+                        -v ${selfContainedVol}":${containerRoot}volume" `
+                        $selfContainedImage `
+                        chmod u+x ${containerRoot}volume${platformDirSeparator}test
```
naamunds: Why does this need to be done inside a Docker container?  Couldn't we just run chmod on `${selfContainedVol}${dirSeparator}test`?

test/run-test.ps1
MichaelSimons: For 1.0 and 1.1, the sdk directory always has children - msbuild and projectjson.  Since the 2.0 sdk doesn't have those variations, it is a leaf directory therefore doesn't end with a dirSeparator.  Because of this, the tests were not run against the 2.0 assets.

test/run-test.ps1
MichaelSimons: A Docker volume is being used here versus volume mapping a host directory into the container.  I am not sure how to do this directly so I did it the only way I know how.  Please let me know if you know of a more elegant solution.  Keep in mind this is temporary until the underlying product issue is fixed.

naamunds: This approach is fine, then.

PR #197, opened 2017-02-15
Revert 2B Patch Tuesday changes

rakeshsinghranchi:


naamunds:
Why is this update being reverted?

PR #226, opened 2017-03-08
Changed update-dependencies to support matching on cli version prefix

MichaelSimons:
For the cli 2.0 release there is only one moniker therefore we don't need to match on it.  Since it has been changing from time to time, this change will not require us to update the Maestro subscription as often.

cc @dagood


MichaelSimons:
@dotnet-bot test this please

PR #228, opened 2017-03-08
Updating update-dependencies to use CLI 1.0.1

MichaelSimons:
@dagood


MichaelSimons:
IMO we should be dogfooding and showcasing our latest bits.  Since this is standalone, it was very easy to upgrade.

PR #253, opened 2017-03-28
Refactor update-dependencies to update the runtime Dockerfiles

MichaelSimons:
I'm not a fan of this.  I view it as a stop-gap until the produces and consumes build tools infrastructure is in place.

### DIFF
--- a/update-dependencies/CliDependencyHelper.cs
+++ b/update-dependencies/CliDependencyHelper.cs
@@ -1,0 +1,72 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+
+namespace Dotnet.Docker.Nightly
+{
+    public static class CliDependencyHelper
+    {
+        private static readonly Lazy<HttpClient> DownloadClient = new Lazy<HttpClient>();
+
+        public static string GetSharedFrameworkVersion(string cliVersion)
+        {
+            Trace.TraceInformation($"Looking for the Shared Framework CLI '{cliVersion}' depends on.");
+
+            string cliCommitHash = GetCommitHash(cliVersion);
+            XDocument depVersions = DownloadDependencyVersions(cliCommitHash).Result;
+            XNamespace msbuildNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
+            string sharedFrameworkVersion = depVersions.Document.Root
+                .Element(msbuildNamespace + "PropertyGroup")
+                ?.Element(msbuildNamespace + "CLI_SharedFrameworkVersion")
+                ?.Value;
+            if (sharedFrameworkVersion == null)
+            {
+                throw new InvalidOperationException("Can't find CLI_SharedFrameworkVersion in DependencyVersions.props.");
+            }
+
+            Trace.TraceInformation($"Detected Shared Framework version '{sharedFrameworkVersion}'.");
+            return sharedFrameworkVersion;
+        }
+
+        private static string GetCommitHash(string cliVersion)
+        {
+            using (ZipArchive archive = DownloadCliInstaller(cliVersion).Result)
+            {
+                ZipArchiveEntry versionTxtEntry = archive.GetEntry($"sdk/{cliVersion}/.version");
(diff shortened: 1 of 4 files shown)


update-dependencies/Program.cs
```diff
@@ -63,25 +65,35 @@ private static bool ParseArgs(string[] args)
...
+
+            IEnumerable<DependencyBuildInfo> buildInfos = new[]
+                {
+                    new DependencyBuildInfo(cliBuildInfo, false, Enumerable.Empty<string>()),
+                    new DependencyBuildInfo(
+                        new BuildInfo() {
+                            Name = SharedFrameworkBuildInfoName,
+                            LatestReleaseVersion = sharedFrameworkVersion,
+                            LatestPackages = new Dictionary<string, string>()
+                            },
```
dagood: \<nit> Unindent?

update-dependencies/Program.cs
```diff
@@ -98,11 +110,12 @@ private static Task CreatePullRequest(DependencyUpdateResults updateResults)
 
         private static IEnumerable<IDependencyUpdater> GetUpdaters()
         {
-            return Directory.GetFiles(s_repoRoot, "Dockerfile", SearchOption.AllDirectories)
-                .Select(path => CreateDependencyUpdater(path));
+            string[] dockerfiles = Directory.GetFiles(s_repoRoot, "Dockerfile", SearchOption.AllDirectories);
+            return dockerfiles.Select(path => CreateSdkUpdater(path))
+                .Union(dockerfiles.Select(path => CreateRuntimeUpdater(path)));
```
dagood: Isn't `Concat` enough?

update-dependencies/update-dependencies.csproj
```diff
@@ -6,7 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="1.0.26-prerelease-00708-05" />
+    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="1.0.27-prerelease-01428-01" />
     <PackageReference Include="System.Diagnostics.TextWriterTraceListener" Version="4.0.0" />
+    <PackageReference Include="NuGet.Packaging.Core.Types" Version="4.0.0" />
+    <PackageReference Include="NuGet.Packaging" Version="4.0.0" />
```
dagood: Why were the NuGet packages added?

update-dependencies/CliDependencyHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        private static readonly Lazy<HttpClient> DownloadClient = new Lazy<HttpClient>();
+
+        public static string GetSharedFrameworkVersion(string cliVersion)
+        {
+            Trace.TraceInformation($"Looking for the Shared Framework CLI '{cliVersion}' depends on.");
+
+            string cliCommitHash = GetCommitHash(cliVersion);
+            XDocument depVersions = DownloadDependencyVersions(cliCommitHash).Result;
+            XNamespace msbuildNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
```
dagood: I usually get the namespace of the `XDocument` for flexibility, and I might use `Descendants` below, but it's probably not needed here.

update-dependencies/Program.cs
MichaelSimons: Fixed.

update-dependencies/Program.cs
MichaelSimons: Yes - changed.

update-dependencies/CliDependencyHelper.cs
MichaelSimons: Good point on the namespace - changed.
It seems more precise to use Elements with an explicit path - leaving.

update-dependencies/update-dependencies.csproj
MichaelSimons: Picking up the new versions package required 4.0.0 of NuGet.  Without, it is trying to use 3.5.0.  I was able to remove the NuGet.Packaging.Core.Types reference.

MichaelSimons: I should reword, without it it is causing a build warning - warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.

dagood: Looks like the NuGet.Packaging dependency version in the VersionTools nuspec didn't get updated when the one in the VersionTools project.json did (annoying duplication there)... which means this fix makes sense, but it seems like this should really be fixed in BuildTools.

PR #300, opened 2017-05-16
Misc CI script cleanup

MichaelSimons:
- Updated CI scripts to read Docker Hub repo info from manifest.  This makes it easier to test changes when you want to build and push to a private repo.
- Disabled the NuGet package workaround for when packages aren't yet available on NuGet.org.

### DIFF
--- a/test/run-test.ps1
+++ b/test/run-test.ps1
@@ -20,9 +20,9 @@ else {
     $optionalDockerBuildArgs = "--no-cache"
 }
 
-$dockerRepo = "microsoft/dotnet-nightly"
 $dirSeparator = [IO.Path]::DirectorySeparatorChar
 $repoRoot = Split-Path -Parent $PSScriptRoot
+$dockerRepo = (Get-Content "${repoRoot}${dirSeparator}manifest.json" | ConvertFrom-Json).DockerRepo
 $testFilesPath = "$PSScriptRoot$dirSeparator"
 $platform = docker version -f "{{ .Server.Os }}"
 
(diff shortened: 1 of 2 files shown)


test/run-test.ps1
```diff
@@ -20,9 +20,9 @@ else {
     $optionalDockerBuildArgs = "--no-cache"
 }
 
-$dockerRepo = "microsoft/dotnet-nightly"
 $dirSeparator = [IO.Path]::DirectorySeparatorChar
 $repoRoot = Split-Path -Parent $PSScriptRoot
+$dockerRepo = (Get-Content "${repoRoot}${dirSeparator}manifest.json" | ConvertFrom-Json).DockerRepo
```
dagood: Do `[IO.Path]::Combine` or `Join-Path` work here?

test/run-test.ps1
```diff
@@ -85,10 +85,10 @@ Get-ChildItem -Path $repoRoot -Recurse -Filter Dockerfile |
...
                 $selfContainedImage = "self-contained-build-${buildImage}"
                 $optionalRestoreParams = ""
-                if ($sdkTag -like "2.0-sdk") {
-                    # Temporary workaround until 2.0 packages are released on NuGet.org
-                    $optionalRestoreParams = "-s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json -s https://api.nuget.org/v3/index.json"
-                }
+                # Enable and update as appropriate whenever pre-release packages are referenced prior to being available on NuGet.org.
+                # if ($sdkTag -like "<???>-sdk") {
+                #     $optionalRestoreParams = "-s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json -s https://api.nuget.org/v3/index.json"
+                # }
```
dagood: I think the diffs would be cleaner if there were a `$includePrereleasePackageSourceForSdkTag = $null` (or something) conditioned on instead. That var could also be put up at the top of the file. (Although that would be more useful if there were multiple manual tweaks to gather together.)

test/run-test.ps1
MichaelSimons: changed to 
$manifestPath = [IO.Path]::combine(${repoRoot}, "manifest.json")
$dockerRepo = (Get-Content $manifestPath | ConvertFrom-Json).DockerRepo

dagood: When I wrote that, I was thinking it could replace `$dirSeparator`, but I see now it's used for things that combine/join can't do. I'm fine using `$dirSeparator` here, too--either way.

MichaelSimons: I kind of figured that 😃.

test/run-test.ps1
MichaelSimons: I'm not real fond of this pattern to begin with.  I see opportunity to clean some of this up once multi-stage builds are available.  I changed as suggested.

MichaelSimons:
@dotnet-bot test NanoServer please.

MichaelSimons:
@dotnet-bot test NanoServer please.

PR #315, opened 2017-06-01
Update 2.0 Dockerfiles to include checksum verification

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/43

### DIFF
--- a/update-dependencies/DockerfileShaUpdater.cs
+++ b/update-dependencies/DockerfileShaUpdater.cs
@@ -1,0 +1,68 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.VersionTools;
+using Microsoft.DotNet.VersionTools.Dependencies;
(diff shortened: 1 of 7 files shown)


update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -0,0 +1,68 @@
...
+                        + shaExt;
+
+                    Trace.TraceInformation($"Downloading '{shaUrl}'.");
+                    using (Stream shaStream = new HttpClient().GetStreamAsync(shaUrl).Result)
+                    using (StreamReader reader = new StreamReader(shaStream))
+                    {
+                        sha = reader.ReadToEnd();
+                    }
+                }
+            }
```
dagood: Would it be useful to log something if the regexes don't match? A silent NOP might be hard to diagnose, and since you have to point this updater to a specific file I think it makes sense to make a little noise. Maybe it'd be obvious enough from the diff and test output.

MichaelSimons: It certainly won't hurt - added

PR #317, opened 2017-06-08
Add debian:stretch based Dockerfiles for 2.0

MichaelSimons:
1. Added new Debian Stretch based Dockerfiles for 2.0
2. Updated CI scripts to read manifest in order to determine what to build/test.  I would eventually like to remove the build script and replace it with the image-builder tool but that is currently blocked on some Windows/Docker issues.

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -6,9 +6,6 @@ param(
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
 
-$dirSeparator = [IO.Path]::DirectorySeparatorChar
-$dockerRepo = (Get-Content "manifest.json" | ConvertFrom-Json).Repos[0].Name
-
 if ($UseImageCache) {
     $optionalDockerBuildArgs = ""
 }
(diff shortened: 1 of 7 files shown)


build-and-test.ps1
```diff
@@ -6,48 +6,45 @@ param(
...
-        docker build $optionalDockerBuildArgs -t $tag $_.DirectoryName
-        if (-NOT $?) {
-            throw "Failed building $tag"
-        }
+$builtTags = [System.Collections.ArrayList]@()
+
+$manifestRepo.Images |
+    ForEach-Object {
+        $images = $_
+        ForEach-Object {$_.Platforms} |
```
dagood: This usage is very strange to me. Why not just `$_.Platforms |`? As far as I can tell, in this situation (`ForEach-Object` not given any inputs) `$_` keeps its value from the outside scope.

MichaelSimons: Your right this was a leftover from refactoring. Removed `ForEach-Object`

build-and-test.ps1
```diff
@@ -6,48 +6,45 @@ param(
...
+            ForEach-Object {
+                $dockerfilePath = $_.$platform.dockerfile
+                $tags = $_.$platform.Tags
+                if ([bool]($images.PSobject.Properties.name -match "sharedTags")) {
+                    $tags += $images.sharedTags
+                }
+
+                $qualifiedTags = $tags | ForEach-Object {
+                    $_ = $manifestRepo.Name + ':' + $_.Replace('$(nanoServerVersion)', $manifest.TagVariables.NanoServerVersion)
+                    $_
```
MichaelSimons: No reason - fixed.

build-and-test.ps1
```diff
@@ -6,48 +6,45 @@ param(
...
-                Replace("$PSScriptRoot$dirSeparator", '').
-                Replace("$dirSeparator$imageOs", '').
-                Replace($dirSeparator, '-')
-        $tags.Add($tag) | Out-Null
-        Write-Host "--- Building $tag from $($_.DirectoryName) ---"
-        docker build $optionalDockerBuildArgs -t $tag $_.DirectoryName
-        if (-NOT $?) {
-            throw "Failed building $tag"
-        }
+$builtTags = [System.Collections.ArrayList]@()
```
MichaelSimons: No reason - changed to array.

PR #335, opened 2017-06-27
Update manifest.json to tag images as preview3

natemcmaster:
Current nightly builds are tagged as preview2 but contain preview3 assets.


natemcmaster:
Nvm, this is redundant

PR #352, opened 2017-07-10
Break apart builds into multiple legs to reduce time

MichaelSimons:
In the test build I ran the longest windows build leg took 12.9 minutes whereas the single Linux build leg took 14.6 minutes.


MichaelSimons:
Related to https://github.com/dotnet/docker-tools/issues/24

MichaelSimons:
Decided there is a lot of benefit to do this for Linux as well.

PR #369, opened 2017-07-20
Remove test workaround for self-contained apps

MichaelSimons:
Related dotnet/dotnet-docker#270


MichaelSimons:
@dagood - can you take a look?

PR #390, opened 2017-08-07
Remove '--command' arg from builds

MichaelSimons:
This change is needed with the latest image-builder as commands are now parameters.


MichaelSimons:
Merging to fix broken builds.

Issue #364, opened 2017-07-13
Update 1.x images to use latest runtime and SDK nightlies


natemcmaster:
The latest nightly build of the 1.1 SDK is 1.1.0-preview1-005089 and latest 1.1 runtime is 1.1.3-servicing-001554-00, but the dotnet-nightly images still use the stable releases.

Latest SDK: https://github.com/dotnet/cli/tree/rel/1.1.0#installers-and-binaries
Latest runtime: https://github.com/dotnet/core-setup#daily-builds

MichaelSimons:
FYI - Latest SDK has not picked up 1.1.3 runtime yet - https://github.com/dotnet/cli/blob/rel/1.1.0/build/Microsoft.DotNet.Cli.DependencyVersions.props#L4

PR #395, opened 2017-08-08
Update to sdk 1.1.0

MichaelSimons:
Related #364 

Note:  Latest SDK hasn't picked up the latest runtime versions yet therefore the runtime images won't be updated at this time.

### DIFF
--- a/1.0/sdk/jessie/Dockerfile
+++ b/1.0/sdk/jessie/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Install .NET Core SDK
-ENV DOTNET_SDK_VERSION 1.0.4
+ENV DOTNET_SDK_VERSION 1.1.0-preview1-005124
 ENV DOTNET_SDK_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$DOTNET_SDK_VERSION/dotnet-dev-debian-x64.$DOTNET_SDK_VERSION.tar.gz
 
 RUN curl -SL $DOTNET_SDK_DOWNLOAD_URL --output dotnet.tar.gz \
--- a/update-dependencies/CliDependencyHelper.cs
+++ b/update-dependencies/CliDependencyHelper.cs
@@ -11,15 +11,29 @@ using System.Xml.Linq;
 
 namespace Dotnet.Docker.Nightly
 {
-    public static class CliDependencyHelper
+    public class CliDependencyHelper
     {
         private static readonly Lazy<HttpClient> DownloadClient = new Lazy<HttpClient>();
 
-        public static string GetSharedFrameworkVersion(string cliVersion)
-        {
-            Trace.TraceInformation($"Looking for the Shared Framework CLI '{cliVersion}' depends on.");
+        public string CliVersion { get; }
+        public int CliMajorVersion { get; }
 
-            string cliCommitHash = GetCommitHash(cliVersion);
+        public CliDependencyHelper(string cliVersion)
+        {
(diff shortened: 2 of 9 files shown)


update-dependencies/CliDependencyHelper.cs
```diff
@@ -35,11 +49,12 @@ public static string GetSharedFrameworkVersion(string cliVersion)
...
         }
 
-        private static string GetCommitHash(string cliVersion)
+        private string GetCommitHash()
         {
-            using (ZipArchive archive = DownloadCliInstaller(cliVersion).Result)
+            using (ZipArchive archive = DownloadCliInstaller().Result)
             {
-                ZipArchiveEntry versionTxtEntry = archive.GetEntry($"sdk\\{cliVersion}\\.version");
+                string zipEntryName = CliMajorVersion == 1 ? $"sdk/{CliVersion}/.version" : $"sdk\\{CliVersion}\\.version";
```
crummel: I think you should be able to use "sdk/{CliVersion}/.version" in all cases and ZipArchive will handle it (this is the behavior I saw on the symbol signing stuff at least).

update-dependencies/Program.cs
```diff
@@ -117,16 +118,33 @@ private static Task CreatePullRequest(DependencyUpdateResults updateResults)
...
             string majorMinorVersion = s_config.CliReleasePrefix.Substring(0, s_config.CliReleasePrefix.LastIndexOf('.'));
-            string searchFolder = Path.Combine(s_repoRoot, majorMinorVersion);
-            string[] dockerfiles = Directory.GetFiles(searchFolder, "Dockerfile", SearchOption.AllDirectories);
+            string[] dockerfiles = GetDockerfiles(majorMinorVersion);
             Trace.TraceInformation("Updating the following Dockerfiles:");
             Trace.TraceInformation($"{string.Join(Environment.NewLine, dockerfiles)}");
-            return dockerfiles.Select(path => CreateDockerfileEnvUpdater(path, "DOTNET_SDK_VERSION", CliBuildInfoName))
-                .Concat(dockerfiles.Select(path => CreateDockerfileEnvUpdater(path, "DOTNET_VERSION", SharedFrameworkBuildInfoName)))
-                .Concat(dockerfiles.Select(path => new DockerfileShaUpdater(path)));
+            IEnumerable<IDependencyUpdater> updators = dockerfiles
```
crummel: Spelling: "updaters".

1.0/sdk/jessie/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Install .NET Core SDK
-ENV DOTNET_SDK_VERSION 1.0.4
+ENV DOTNET_SDK_VERSION 1.1.0-preview1-005124
```
MichaelSimons: The .NET Core docker versioning is in terms of the Runtime.  Think of it this way, if you want to build an app that targets 1.0, then this is the SDK image you should use to build your app.

update-dependencies/Program.cs
MichaelSimons: Fixed.

update-dependencies/CliDependencyHelper.cs
MichaelSimons: Looking at the [ZipArchive.GetEntry implementation](https://github.com/dotnet/corefx/blob/2f554b13fe67c8cf9b37c70a711de1ba5575638c/src/System.IO.Compression/src/System/IO/Compression/ZipArchive.cs#L261), a simple Dictionary.TryGetValue is used and no transformation is performed on the FullName.  This leads me to conclude the path syntax must match the way the zip was constructed.

Issue #398, opened 2017-08-11
Add 2.1 Dockerfiles


MichaelSimons:
2.1 Dockerfiles should be created

PR #400, opened 2017-08-11
Add 2.1 Dockerfiles

MichaelSimons:
Related #398

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,12 +10,18 @@ See [dotnet/dotnet-docker](https://github.com/dotnet/dotnet-docker) for images w
 - [`1.1.2-runtime-jessie`, `1.1.2-runtime`, `1.1-runtime`, `1-runtime` (*1.1/runtime/jessie/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/1.1/runtime/jessie/Dockerfile)
 - [`1.1.2-runtime-deps-jessie`, `1.1.2-runtime-deps`, `1.1-runtime-deps`, `1-runtime-deps` (*1.1/runtime-deps/jessie/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/1.1/runtime-deps/jessie/Dockerfile)
 - [`1.1.2-sdk-jessie`, `1.1.2-sdk`, `1.1-sdk`, `1-sdk` (*1.1/sdk/jessie/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/1.1/sdk/jessie/Dockerfile)
-- [`2.0.0-runtime-stretch`, `2.0.0-runtime`, `2.0-runtime`, `2-runtime`, `runtime` (*2.0/runtime/stretch/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/runtime/stretch/amd64/Dockerfile)
-- [`2.0.0-runtime-jessie`, `2.0-runtime-jessie`, `2-runtime-jessie` (*2.0/runtime/jessie/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/runtime/jessie/amd64/Dockerfile)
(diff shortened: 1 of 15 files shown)


MichaelSimons:
@dotnet-bot test ci please

README.md
```diff
@@ -10,26 +10,36 @@ See [dotnet/dotnet-docker](https://github.com/dotnet/dotnet-docker) for images w
...
+- [`2.0.0-runtime-jessie`, `2.0-runtime-jessie` (*2.0/runtime/jessie/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/runtime/jessie/amd64/Dockerfile)
+- [`2.0.0-runtime-deps-stretch`, `2.0.0-runtime-deps`, `2.0-runtime-deps` (*2.0/runtime-deps/stretch/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/runtime-deps/stretch/amd64/Dockerfile)
+- [`2.0.0-runtime-deps-jessie`, `2.0-runtime-deps-jessie` (*2.0/runtime-deps/jessie/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/runtime-deps/jessie/amd64/Dockerfile)
+- [`2.0.0-sdk-stretch`, `2.0.0-sdk`, `2.0-sdk` (*2.0/sdk/stretch/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/sdk/stretch/amd64/Dockerfile)
+- [`2.0.0-sdk-jessie`, `2.0-sdk-jessie` (*2.0/sdk/jessie/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/sdk/jessie/amd64/Dockerfile)
+- [`2.1.0-runtime-stretch`, `2.1.0-runtime`, `2.1-runtime`, `2-runtime`, `runtime` (*2.1/runtime/stretch/amd64/Dockerfile*)](https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.1/runtime/stretch/amd64/Dockerfile)
```
natemcmaster: Should we label these 2.1.0-preview1-* for now?

natemcmaster:
By the way, https://github.com/aspnet/aspnet-docker/issues/281 makes me think we should be more careful with previews and the tags we give them.

MichaelSimons:
@natemcmaster - agreed these should be tagged as 2.1.0-preview1-*.  That was an oversight on my part.  That being said, regarding "we should be more careful with previews", how are you suggesting we should be more careful?  In the past, to my recollection we have always indicated "preview" in our fully qualified tags of the official repos.  For example I don't think we used 2.0.0 prior to RTM.  Am I mistaken on this?  The nightly repos are a different story as we of course need to use the RTM tags prior to the actual RTM.

natemcmaster:
> For example I don't think we used 2.0.0 prior to RTM.

Yeah, we're fine there. 

My concern is about the less-than-fully-qualified tags, such as `2.1`, and `2`. Maybe we should avoid adding these tags to previews. They are likely to break between previews. Once we RTM, the tags with `major.minor` should be fairly stable. In fact, these are the tags I recommend so you get patches automatically.

Side note: the `major` tag still leaves open the possibility of breaking when updates are published. A 2.0 app will not work on 2.1. The TFM, runtimes, runtime store, etc. all change too much between minor releases, but this is probably a conversation for another time.

kendrahavens:
@natemcmaster Michael and I synced a little on this. We had understood that the 2.1 and 2 preview tags breaking when updated is actually a benefit so they will notify the CI machine (or user) when they are out of date. This is why we've used this pattern before.

natemcmaster:
I think users would prefer to never break. IMO it would be better if we didn't publish `major.minor` until RTM, and dropped `major` altogether.

MichaelSimons:
When settling on the Docker tagging strategy for .NET Core, the discussions about the `major.minor` and `major` tags centered around developer scenarios.  Imagine a CI system setup to always build on these tags as a way to detect breaks.

That being said, we are certainly open to feedback from the community on the tags we produce.  Having not received any community feedback thus far, I am reluctant to make changes.

MichaelSimons:
Going to merge in order to get 2.1 images.  If there are are further suggestions/imput on tagging please open an issue.  We have time to make changes before the 2.1 images are migrated to microsoft/dotnet.

PR #415, opened 2017-09-08
Refactor Powershell based tests as xunit tests

MichaelSimons:
The Powershell tests are just getting too complicated to maintain.  Trying xunit tests instead using C#.  This is needed as part of https://github.com/dotnet/dotnet-docker/issues/286 which requires more complicated logic to determine which sdk, runtime-deps, runtime image pairings to test together as each image will not exist for every version.

### DIFF
--- a/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -11,9 +11,9 @@ namespace Microsoft.DotNet.Docker.Tests
 {
     public class DockerHelper
     {
-        public string DockerOS => GetDockerOS();
-        public string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
-        public bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
+        public static string DockerOS => GetDockerOS();
+        public static string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
+        public static bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
         private ITestOutputHelper Output { get; set; }
 
         public DockerHelper(ITestOutputHelper output)
--- a/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -11,7 +11,7 @@ namespace Microsoft.DotNet.Docker.Tests
 {
     public class DockerHelper
     {
-        private string DockerOS => GetDockerOS();
+        public string DockerOS => GetDockerOS();
         public string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
         public bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
         private ITestOutputHelper Output { get; set; }
@@ -24,13 +24,13 @@ namespace Microsoft.DotNet.Docker.Tests
         public void Build(string dockerfilePath, string fromImage, string tag, string buildArgs)
         {
             string dockerfileContents = File.ReadAllText(dockerfilePath);
-            dockerfileContents = dockerfileContents.Replace("$base_image", fromImage);
+            dockerfileContents = dockerfileContents.Replace("{base_image}", fromImage);
             string tempDockerfilePath = dockerfilePath + ".temp";
             File.WriteAllText(tempDockerfilePath, dockerfileContents);
 
             try
             {
-                Execute($"build -t {tag} {buildArgs} -f {dockerfilePath} .");
+                Execute($"build -t {tag} {buildArgs} -f {tempDockerfilePath} .");
             }
             finally
             {
--- a/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -25,8 +25,8 @@ namespace Microsoft.DotNet.Docker.Tests
         {
             string dockerfileContents = File.ReadAllText(dockerfilePath);
             dockerfileContents = dockerfileContents.Replace("$base_image", fromImage);
-            dockerfilePath = dockerfilePath + ".temp";
-            File.WriteAllText(dockerfilePath, dockerfileContents);
+            string tempDockerfilePath = dockerfilePath + ".temp";
+            File.WriteAllText(tempDockerfilePath, dockerfileContents);
 
             try
             {
@@ -34,7 +34,7 @@ namespace Microsoft.DotNet.Docker.Tests
             }
             finally
             {
-                File.Delete(dockerfilePath);
+                File.Delete(tempDockerfilePath);
             }
         }
 
--- a/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -1,0 +1,105 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class DockerHelper
+    {
+        private string DockerOS => GetDockerOS();
+        public string ContainerWorkDir => IsLinuxContainerModeEnabled ? "/sandbox" : "c:\\sandbox";
+        public bool IsLinuxContainerModeEnabled => string.Equals(DockerOS, "linux", StringComparison.OrdinalIgnoreCase);
+        private ITestOutputHelper Output { get; set; }
+
+        public DockerHelper(ITestOutputHelper output)
+        {
+            Output = output;
+        }
+
+        public void Build(string dockerfilePath, string fromImage, string tag, string buildArgs)
+        {
+            string dockerfileContents = File.ReadAllText(dockerfilePath);
+            dockerfileContents = dockerfileContents.Replace("$base_image", fromImage);
+            dockerfilePath = dockerfilePath + ".temp";
+            File.WriteAllText(dockerfilePath, dockerfileContents);
+
+            try
+            {
+                Execute($"build -t {tag} {buildArgs} -f {dockerfilePath} .");
+            }
+            finally
+            {
+                File.Delete(dockerfilePath);
+            }
+        }
+
+        public void DeleteImage(string name)
+        {
+            if (ResourceExists("image", name))
+            {
+                Execute($"image rm -f {name}");
+            }
+        }
+
+        public void DeleteVolume(string name)
+        {
+            if (ResourceExists("volume", name))
+            {
+                Execute($"volume rm -f {name}");
+            }
+        }
+
+        private void Execute(string args)
+        {
+            Output.WriteLine($"Executing : docker {args}");
+            ProcessStartInfo info = new ProcessStartInfo("docker", args);
+            info.RedirectStandardOutput = true;
+            info.RedirectStandardError = true;
+            Process process = Process.Start(info);
+            process.WaitForExit();
+            Output.WriteLine(process.StandardOutput.ReadToEnd());
+
+            if (process.ExitCode != 0)
+            {
+                string stdErr = process.StandardError.ReadToEnd();
+                string msg = $"Failed to execute {info.FileName} {info.Arguments}{Environment.NewLine}{stdErr}";
+                throw new InvalidOperationException(msg);
+            }
+        }
+
+        private static string GetDockerOS()
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo("docker", "version -f \"{{ .Server.Os }}\"");
+            startInfo.RedirectStandardOutput = true;
+            Process process = Process.Start(startInfo);
+            process.WaitForExit();
+            return process.StandardOutput.ReadToEnd().Trim();
+        }
+
+        public string GetContainerWorkPath(string relativePath)
+        {
+            string separator = IsLinuxContainerModeEnabled ? "/" : "\\";
+            return $"{ContainerWorkDir}{separator}{relativePath}";
+        }
+
+        private static bool ResourceExists(string type, string id)
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo("docker", $"{type} ls -q {id}");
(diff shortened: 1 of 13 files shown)


test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+            ProcessStartInfo startInfo = new ProcessStartInfo("docker", "version -f \"{{ .Server.Os }}\"");
+            startInfo.RedirectStandardOutput = true;
+            Process process = Process.Start(startInfo);
+            process.WaitForExit();
+            return process.StandardOutput.ReadToEnd().Trim();
+        }
+
+        public string GetContainerWorkPath(string relativePath)
+        {
+            string separator = IsLinuxContainerModeEnabled ? "/" : "\\";
```
ravimeda: Instead of `"\\"` why not `$"{Path.DirectorySeparatorChar}"`

test/run-test.ps1
```diff
@@ -1,169 +1,62 @@
...
-}
-
-function Get-ActivePlatformImages([PSCustomObject]$manifestRepo, [string]$activeOS) {
-    return $manifestRepo.Images |
-        ForEach-Object { $_.Platforms } |
-        Where-Object { $_.os -eq "$activeOS" } |
-        Where-Object { ( [string]::IsNullOrEmpty($Architecture) -and -not [bool]($_.PSobject.Properties.name -match "architecture"))`
-            -or ( [bool]($_.PSobject.Properties.name -match "architecture") -and $_.architecture -eq "$Architecture" ) }
-}
+$DotnetInstallDir = "$PSScriptRoot/../.dotnet"
```
ravimeda: Why is `DotnetInstallDir` a parameter when it gets initialized this way?

test/run-test.ps1
```diff
@@ -1,169 +1,62 @@
...
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
 [cmdletbinding()]
 param(
-    [switch]$UseImageCache,
+    [string]$DotnetInstallDir,
     [string]$Filter,
     [string]$Architecture
```
ravimeda: We could default `Architecture` here as `[string]$Architecture="amd64"`

test/run-test.ps1
```diff
@@ -1,169 +1,62 @@
...
-$manifestPath = [IO.Path]::combine(${repoRoot}, "manifest.json")
-$manifestRepo = (Get-Content $manifestPath | ConvertFrom-Json).Repos[0]
-$testFilesPath = "$PSScriptRoot$dirSeparator"
-$activeOS = docker version -f "{{ .Server.Os }}"
-
-# update as appropriate (e.g. "2.0-sdk") whenever pre-release packages are referenced prior to being available on NuGet.org.
-$includePrereleasePackageSourceForSdkTag = "2.*-sdk*"
+if (!(Test-Path $DotnetInstallScript)) {
+    $DOTNET_INSTALL_SCRIPT_URL = "https://raw.githubusercontent.com/dotnet/cli/release/2.0.0/scripts/obtain/$DotnetInstallScript"
+    Invoke-WebRequest $DOTNET_INSTALL_SCRIPT_URL -OutFile $DotnetInstallDir/$DotnetInstallScript
```
ravimeda: Should we check `$LASTEXITCODE` to ensure download succeeded?

test/run-test.ps1
```diff
@@ -1,169 +1,62 @@
...
+    if (![string]::IsNullOrWhiteSpace($Filter)) {
+        $TestFilter += "&Version=$filter"
+    }
 
-                        exec { docker run --rm `
-                            -v ${selfContainedVol}":${containerRoot}volume" `
-                            $selfContainedImage `
-                            dotnet publish -r debian.8-x64 -o ${containerRoot}volume $optionalPublishParams
-                        }
+    & $DotnetInstallDir/dotnet test -v n --filter """$TestFilter"""
```
ravimeda: `""$TestFilter"""` do we need all those quotes?

test/Microsoft.DotNet.Docker.Tests/DotNetImageType.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public enum DotNetImageType
+    {
+        SDK,
```
ravimeda: Minor - should these be alphabetical?

test/run-test.ps1
```diff
@@ -1,169 +1,62 @@
...
-                        $optionalRestoreParams = " /p:runtimeidentifier=debian.8-x64"
-                    }
-
-                    exec { (Get-Content ${testFilesPath}Dockerfile.linux.publish).
-                                Replace("{image}", $buildImage).
-                                Replace("{optionalRestoreParams}", $optionalRestoreParams) `
-                        | docker build $optionalDockerBuildArgs -t $selfContainedImage -
-                    }
+Try {
+    # Run Tests
```
ravimeda: Minor - Precondition to run tests is `Dockerfile.test` is available. If in the unlikely case it isn't available, better to display error and exit.

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+
+        private static string GetBuildArgs(bool usePreReleasePackageFeed, params string[] additionalArgs)
+        {
+            string buildArgs = string.Empty;
+
+            if (usePreReleasePackageFeed)
+            {
+                buildArgs += " --build-arg optional_restore_feeds=\"-s https://dotnet.myget.org/F/dotnet-core/api/v3/index.json -s https://api.nuget.org/v3/index.json\"";
```
ravimeda: Minor - move these constants to top of this file or a separate constants file.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+        public DockerHelper(ITestOutputHelper output)
+        {
+            Output = output;
+        }
+
+        public void Build(string dockerfilePath, string fromImage, string tag, string buildArgs)
+        {
+            string dockerfileContents = File.ReadAllText(dockerfilePath);
+            dockerfileContents = dockerfileContents.Replace("$base_image", fromImage);
+            dockerfilePath = dockerfilePath + ".temp";
```
ravimeda: Instead of `dockerfilePath = dockerfilePath + ".temp";`

`string tempFilePath = dockerfilePath + ".temp";` or similar would make it more clear.

test/run-test.ps1
ravimeda: Better to make it a validate set. For example, `[ValidateSet("x86","amd64","arm")] `

test/run-test.ps1
crummel: I think `Invoke-WebRequest` would be a try-catch instead of `$LASTEXITCODE` since it's internal to PowerShell.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
MichaelSimons: I would prefer to not make an assumption that the separator within the container the not the same as the host OS.

test/run-test.ps1
MichaelSimons: An oversight on my part - removing parameter.

test/run-test.ps1
MichaelSimons: I had it that way to begin with but you can pass "-Architecture" which is exactly what build-and-test.ps1 does.  When this occurs I want to use "amd64".

test/run-test.ps1
MichaelSimons: No - This is a cmdlet.  I will set $ErrorActionPreference to stop to handle this.

test/run-test.ps1
MichaelSimons: $TestFilter needs to be quoted and quotes within quotes need to be escaped.  Escaping can be done with "" or '".

test/Microsoft.DotNet.Docker.Tests/DotNetImageType.cs
MichaelSimons: I'm following the principle that the default enum value should be the most common value.  Since the SDK image is the "latest" image, having the SDK be the default value makes most sense to me.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
ravimeda: Valid point. Thanks!

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+            }
+            finally
+            {
+                File.Delete(dockerfilePath);
+            }
+        }
+
+        public void DeleteImage(string name)
+        {
+            if (ResourceExists("image", name))
```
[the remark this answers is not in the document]
MichaelSimons: Correct - I wish the Docker CLI had support for ignoring non-existent images but they don't.

test/run-test.ps1
ravimeda: Yes, let's keep it the way it was.

test/run-test.ps1
MichaelSimons: The Dockerfile.test gets copied to the output dir as part of the build.  Since dotnet test performs the build, I'm not sure how you would check the precondition.  Did I misunderstand the comment?

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
MichaelSimons: Changed to tempDockerfilePath.  FYI - this code will be removed as soon as the CI machines are upgraded to a version of Docker which supports multi-stage builds.  This will allow the usage of ARGs in the FROM.  It will make this code so much simpler.

test/run-test.ps1
ravimeda: Thanks.  Couple follow up questions -

1. Do we want to wait-retry?
2. Should we get this from Azure and not GitHub?

test/run-test.ps1
ravimeda: Please ignore my comment. I misunderstood how Dockerfile.test is used in build.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
ravimeda: Ok...thanks.

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: hmm - what value would that provide.

ravimeda: Maintainability.  For example, if I were to switch from v3 to vNext feeds, then I would look for a variable or a property corresponding to the feed. Having such variables clustered at the top would make it easier to locate and edit.

I know we can argue that in the current situation, the feeds are defined-used in only one line/statement, and thus a need for a variable is not required. So in this case we can keep it the way it is.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public void DeleteVolume(string name)
+        {
+            if (ResourceExists("volume", name))
+            {
+                Execute($"volume rm -f {name}");
+            }
+        }
+
+        private void Execute(string args)
```
ravimeda: How about an Execute method that can return exit code and output text.

```C#
private (int ExitCode, string StandardOutput) Execute(string args, bool isReturnOutputText = false)
        {
            Output.WriteLine($"Executing : docker {args}");

            ProcessStartInfo info = new ProcessStartInfo("docker", args)
            {
                RedirectStandardOutput = true,
                RedirectStandardError = true
            };

            Process process = Process.Start(info);
            process.WaitForExit();
            int exitCode = process.ExitCode;
            Output.WriteLine(process.StandardOutput.ReadToEnd());

            if (exitCode != 0)
            {
                string stdErr = process.StandardError.ReadToEnd();
                string msg = $"Failed to execute {info.FileName} {info.Arguments}{Environment.NewLine}{stdErr}";
                throw new InvalidOperationException(msg);
            }

            string outputText = string.Empty;

            if (isReturnOutputText)
            {
                outputText = process.StandardOutput.ReadToEnd();
            }

            return (exitCode, outputText);
        }
```

This would give us some flexibility. For example, to get Docker version, we would do -

```C#
Output.WriteLine(Execute("version", true).StandardOutput);
```

For this to work we will take a dependency on System.ValueTuple. Not sure what complexity it would add. But wanted to share the thought...

test/run-test.ps1
MichaelSimons: 1.  In the long run yes - I am going to be doing some additional refactoring soon in which I will address this.
2.  I don't see a reasonable option for this as CLI doesn't consider their install script a build artifact therefore it is only available via GitHub.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+using System;
+using System.Diagnostics;
+using System.IO;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class DockerHelper
+    {
+        public string DockerOS => GetDockerOS();
```
ravimeda: `public static string` ?

MichaelSimons: Changed

test/run-test.ps1
```diff
@@ -1,169 +1,64 @@
...
+
 [cmdletbinding()]
 param(
-    [switch]$UseImageCache,
     [string]$Filter,
     [string]$Architecture
 )
 
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
```
ravimeda: Shouldn't we reset `$ErrorActionPreference` to the state before it was modified? Maybe this script is not intended to be invoked directly by a user/developer.  But `build-and-test.ps1` would be invoked directly, and that means we need to reset the preference.

ravimeda:
New changes look good.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
MichaelSimons: I'm not crazy about the usage of Tuples here.

test/run-test.ps1
MichaelSimons: Per our conversation, ErrorActionPreference is scoped to the script in a non-ISE environment.  I don't think restoring is necessary for this usage.

PR #421, opened 2017-09-13
Consolidate 2.0 & 2.1 build legs on linux

MichaelSimons:
This is necessary in order to stop producing the 2.1-runtime-deps images since they are identical to 2.0.  The 2.1 tests will build a 2.1 self-contained app and verify it will run on the 2.0-runtime-deps image.

Related to https://github.com/dotnet/dotnet-docker/issues/286


MichaelSimons:
@dotnet-bot test ci please

PR #426, opened 2017-09-19
Updates for 1.0.7/1.1.4

MichaelSimons:


ravimeda:
LGTM. Clarification - are all these find-replace done through a script?

MichaelSimons:
These were manually done because the Maestro hook is not working.

PR #432, opened 2017-09-22
Dummy commit.

ravimeda:


ravimeda:
@dotnet-bot test ci please.

MichaelSimons:
@ravimeda - What was the purpose of this PR and why was this PR merged?  I assume it was to test CI.  If that was the case, then there is no need to merge it.  We don't want to pollute the repo with files like this in addition to polluting the history.  Please revert.

ravimeda:
As discussed offline - practice of using `dummy.txt` is to invoke a CI test without touch any source code file. For example, in core-setup (https://github.com/dotnet/core-setup/blob/master/Documentation/dummy.txt). For dotnet/docker repos, we do not want to follow such practice.

@mmitche FYI.

MichaelSimons:
core-setup has a checked in `dummy.txt` to trigger actual product builds.  This is not needed for dotnet/docker.

PR #435, opened 2017-09-25
Refactor all utests to use Theory and MemberData

MichaelSimons:
This should make it easier to maintain the tests as new .NET Core versions are released and support is dropped for older versions.

### DIFF
--- a/test/Microsoft.DotNet.Docker.Tests/VerifyImageDescriptor.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/VerifyImageDescriptor.cs
@@ -1,0 +1,28 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class VerifyImageDescriptor
+    {
+        private string runtimeDepsVersion;
+        private string sdkVersion;
+
+        public string Architecture { get; set; } = "amd64";
+        public string DotNetCoreVersion { get; set; }
+        public string OsVariant { get; set; }
+
+        public string RuntimeDepsVersion
+        {
+            get { return runtimeDepsVersion ?? DotNetCoreVersion; }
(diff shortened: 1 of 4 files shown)


test/run-test.ps1
```diff
@@ -50,12 +50,10 @@ Try {
         $Architecture = "amd64"
     }
 
-    $TestFilter = "Architecture=$Architecture"
-    if (![string]::IsNullOrWhiteSpace($Filter)) {
-        $TestFilter += "&Version~$filter"
-    }
+    $env:IMAGE_ARCH_FILTER = $Architecture
```
ravimeda: Should these environment variable be deleted towards the end of test in a finally block - `Remove-Item Env:\IMAGE_ARCH_FILTER -ErrorAction Ignore`

test/Microsoft.DotNet.Docker.Tests/VerifyImageDescriptor.cs
```diff
@@ -0,0 +1,28 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.DotNet.Docker.Tests
+{
+    public class VerifyImageDescriptor
```
ravimeda: Minor - `ImageDescriptior` seems to be a better name for this class.

ravimeda:
LGTM. Feel free to merge. I'll rebase with these changes in my PR for ARM tests.

ravimeda:
FYI...the CI failure is unrelated to these changes.

test/Microsoft.DotNet.Docker.Tests/VerifyImageDescriptor.cs
MichaelSimons: Good suggestion - changed

test/run-test.ps1
MichaelSimons: I don't think it is necessary - these env vars are only set for the current PS session and not machine wide.

PR #437, opened 2017-09-26
Use ARGs in FROM of test Dockerfiles

MichaelSimons:
This wasn't possible before because the version of docker on the CI machines didn't support referencing ARGs in the FROM instruction.  This change cleans up the code a little and the tests will run a little faster.


test/Microsoft.DotNet.Docker.Tests/Dockerfile.linux.testapp.selfcontained
```diff
@@ -1,4 +1,5 @@
-FROM {base_image}
+ARG base_image
+FROM $base_image
```
ravimeda: Clarification - `$base_image` vs. `${base_image}` what is the purpose of the curly brackets? Is there a risk by not including them. For example, I would prefer to have quotes around a variable that contains a path `"$programfiles%"`.

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -75,22 +75,21 @@ public void VerifyImages(ImageDescriptor imageDescriptor)
...
             string sdkImage = GetDotNetImage(
                 imageDescriptor.SdkVersion, DotNetImageType.SDK, imageDescriptor.OsVariant);
 
             DockerHelper.Build(
                 dockerfile: $"Dockerfile.{DockerHelper.DockerOS.ToLower()}.testapp",
-                fromImage: sdkImage,
                 tag: appSdkImage,
-                buildArgs: buildArgs);
+                fromImage: sdkImage,
+                buildArgs: buildArgs.ToArray());
```
ravimeda: `ToArray()` is that better (size/time) as opposed to passing the list.

ravimeda:
Looks good. Please rebase.

test/Microsoft.DotNet.Docker.Tests/Dockerfile.linux.testapp.selfcontained
MichaelSimons: $base_image is they way you [reference an ARG within a Dockerfile](https://docs.docker.com/engine/reference/builder/#using-arg-variables).  Before this change {} was the convention we used to search for strings to replace prior to building the Dockerfile.  Your concern on spaces in not applicable as the FROM image cannot contain spaces.

ravimeda: Got it. Thanks.

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: buildArgs is a params therefore it has to be an array

PR #463, opened 2017-10-12
Add escape directive in Nano Server Dockerfiles

ravimeda:
Update Nano Server Dockerfile to use the escape directive (https://docs.docker.com/engine/reference/builder/#escape).

Tested these changes by running `build-and-test.ps1 -Filter *nanoserver*`. Tested the scope of the escape directive through the following sequence -

1. Build an image using a Dockerfile that uses backtick as the escape directive. Say, image tag is `backtick:latest`
2. Prepare a Dockerfile that uses `backtick:latest`. Add some additional commands in the Dockerfile that use `\` backslash as the line continuation character.
3. Build an image using the Dockerfile created in step #2. Expectation is a Dockerfile parse error due to `\` character.

### DIFF
--- a/1.0/runtime/nanoserver/Dockerfile
+++ b/1.0/runtime/nanoserver/Dockerfile
@@ -1,3 +1,5 @@
+# escape=`
+
 FROM microsoft/nanoserver:10.0.14393.1770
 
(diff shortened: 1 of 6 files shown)


MichaelSimons:
Related to #311.

PR #467, opened 2017-10-18
Add nanoserver Dockerfiles for 1709 release

MichaelSimons:


ravimeda:
Minor - Branding is `Nano Server` and not `nanoserver`.

2.1/runtime/nanoserver-1709/amd64/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+FROM microsoft/windowsservercore:1709 AS installer-env
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+# Retrieve .NET Core Runtime
+ENV DOTNET_VERSION 2.1.0-preview1-25816-02
+ENV DOTNET_DOWNLOAD_URL https://dotnetcli.blob.core.windows.net/dotnet/Runtime/$DOTNET_VERSION/dotnet-runtime-$DOTNET_VERSION-win-x64.zip
+ENV DOTNET_DOWNLOAD_SHA 1F7BB23909CBBEA1A010773FD9E7BFE9944E551FEB7DCF10AC5D923901BD13BEB09A2677D528F16398329FD892B4B2975404D842581475C703633C99DD069CC5
+
+RUN Invoke-WebRequest $Env:DOTNET_DOWNLOAD_URL -OutFile dotnet.zip; `
```
ravimeda: Suggest using `UseBasicParsing` parameter for `Invoke-WebRequest` cmdlet

2.1/runtime/nanoserver-1709/amd64/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+    Expand-Archive dotnet.zip -DestinationPath dotnet; `
+    Remove-Item -Force dotnet.zip
+
+
+# Runtime image
+FROM microsoft/nanoserver:1709
+
+COPY --from=installer-env ["dotnet", "C:\\Program Files\\dotnet"]
+
+RUN setx PATH "%PATH%;C:\Program Files\dotnet"
```
ravimeda: I think it is correct, but clarifying - should backslash in `C:\Program Files\dotnet` be escaped?

build-and-test.ps1
```diff
@@ -22,12 +23,17 @@ $manifestRepo = $manifest.Repos[0]
 $activeOS = docker version -f "{{ .Server.Os }}"
 $builtTags = @()
 
+$buildFilter = $Filter
+if ($activeOS -eq "windows" -and $buildFilter -like "2.*") {
```
ravimeda: `-eq` should we do a `-ieq` instead?

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -68,10 +84,19 @@ public static IEnumerable<object[]> GetVerifyImagesData()
                 versionFilterPattern = "^" + Regex.Escape(VersionFilter).Replace(@"\*", ".*").Replace(@"\?", ".") + "$";
```
ravimeda: Consider moving this to a method, which accepts the string to be escaped.
```C#
"^" + Regex.Escape(VersionFilter).Replace(@"\*", ".*").Replace(@"\?", ".") + "$"
```

test/run-test.ps1
```diff
@@ -51,6 +52,7 @@ Try {
     }
 
     $env:IMAGE_ARCH_FILTER = $Architecture
```
ravimeda: Are these alphabetical as in `Architecture`, `OS` and `Version`?

MichaelSimons: These sorted by priority.  Which is most likely to be used.

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: Good suggestion - fixed.

build-and-test.ps1
MichaelSimons: fixed

MichaelSimons: wait - what is the benefit of `-ieq` over `eq`?  Are they equivalent?

2.1/runtime/nanoserver-1709/amd64/Dockerfile
MichaelSimons: Escaping is only necessary with the `[ ]`  syntax.  BTW the combination of our builds and utests will catch issues here.  If this syntax were incorrect we couldn't do things like `RUN dotnet new`

2.1/runtime/nanoserver-1709/amd64/Dockerfile
MichaelSimons: I logged this as a separate issue as I think it should be considered and applied to all of the Windows based Dockerfiles not just the ones I am adding here - https://github.com/dotnet/dotnet-docker-nightly/issues/469

MichaelSimons:
@ravimeda - Can you clarify what you meant by the following?

> Minor - Branding is Nano Server and not nanoserver.

Is that in reference to a specific source change I made or is it in reference to the title of the PR?

ravimeda:
Only in the PR title. All other occurrences can be ignored (or does not apply).

---
In reply to: [337706841](https://github.com/dotnet/dotnet-docker-nightly/pull/467#issuecomment-337706841) [](ancestors = 337706841)

2.1/runtime/nanoserver-1709/amd64/Dockerfile
ravimeda: Thanks!

---
In reply to: [145521279](https://github.com/dotnet/dotnet-docker-nightly/pull/467#discussion_r145521279) [](ancestors = 145521279)

2.1/runtime/nanoserver-1709/amd64/Dockerfile
ravimeda: Thanks.

---
In reply to: [145520154](https://github.com/dotnet/dotnet-docker-nightly/pull/467#discussion_r145520154) [](ancestors = 145520154)

build-and-test.ps1
ravimeda: Case-insensitive version of `-eq`. In case Docker return `Windows` instead of the current `windows`.

---
In reply to: [145519453](https://github.com/dotnet/dotnet-docker-nightly/pull/467#discussion_r145519453) [](ancestors = 145519453)

test/run-test.ps1
ravimeda: Thanks.

---
In reply to: [145518236](https://github.com/dotnet/dotnet-docker-nightly/pull/467#discussion_r145518236) [](ancestors = 145518236)

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -25,41 +26,56 @@ public ImageTests(ITestOutputHelper output)
...
+                testData = new List<ImageDescriptor>
+                {
+                    new ImageDescriptor { DotNetCoreVersion = "1.0", SdkVersion = "1.1" },
+                    new ImageDescriptor { DotNetCoreVersion = "1.1", RuntimeDepsVersion = "1.0" },
+                    new ImageDescriptor { DotNetCoreVersion = "2.0" },
+                    new ImageDescriptor { DotNetCoreVersion = "2.0", OsVariant = "jessie" },
+                    new ImageDescriptor
+                    {
+                        DotNetCoreVersion = "2.0",
+                        OsVariant = "stretch",
```
ravimeda: How about an enumeration? Benefit is allows user to avoid spell errors such `strecth` or `jesse`. Similarly for PlatformOS.

build-and-test.ps1
MichaelSimons: `eq` is by default case-insensitive.  It is my understanding that `ieq` is only necessary if you want to explicitly call out the inequality.  IMO `eq` is sufficient here.

see [https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-5.1](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-5.1)
> By default, all comparison operators are case-insensitive. To make a comparison operator case-sensitive, precede the operator name with a c. For example, the case-sensitive version of -eq is -ceq. To make the case-insensitivity explicit, precede the operator with an i. For example, the explicitly case-insensitive version of -eq is -ieq.

---
In reply to: [145525177](https://github.com/dotnet/dotnet-docker-nightly/pull/467#discussion_r145525177) [](ancestors = 145525177,145519453)

MichaelSimons:
> Only in the PR title. All other occurrences can be ignored (or does not apply).
 I used `nanoserver` as it is the OS's tag name.

build-and-test.ps1
ravimeda: Ok. Let's leave it as-is.

---
In reply to: [145526838](https://github.com/dotnet/dotnet-docker-nightly/pull/467#discussion_r145526838) [](ancestors = 145526838,145525177,145519453)

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: I have some thoughts on how to clean this up in general.  One problem I have with using an enum is that the length increases which doesn't allow for a single line definition which really hurts the readability.  I will make a follow up PR to address this.

ravimeda: I knew you would :) Thanks!

Issue #491, opened 2017-11-14
Align 1.* folder structure with 2.*


ravimeda:
Filtering logic uses a wildcard to match any character(s) after the specified, if any, version and OS filters. 

```PowerShell
$buildFilter = "$VersionFilter*/*/$OSFilter*"
```
Refer https://github.com/dotnet/dotnet-docker-nightly/pull/490/files#diff-5eb43610b02eda87c0c4550f32a64b09R26

This logic was written this way since 1.* folder structure does not align with 2.*. After alignment, this logic can be cleaned up as follows -

```PowerShell
$buildFilter = ''
if (-not [string]::IsNullOrEmpty($versionFilter))
{
$buildFilter = "$versionFilter/$buildFilter"
}
if (-not [string]::IsNullOrEmpty($OsFilter))
{
$buildFilter += "$buildFilter/$OsFilter/"
}
```

PR #496, opened 2017-11-20
Add architecture folder to all versions

MichaelSimons:
Related to #491


MichaelSimons:
@ravimeda - No problem - thanks

PR #509, opened 2018-01-03
Ensure volumes get cleaned up and test data refactor

MichaelSimons:
1. The logic to cleanup Docker volumes was failing.  A change was required to correctly check if the volume existed or not.
2. I refactored how the test data is declared slightly to hopefully make it easier to maintain going forward.

### DIFF
--- a/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
+++ b/test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
@@ -44,7 +44,7 @@ namespace Microsoft.DotNet.Docker.Tests
 
         public void DeleteVolume(string name)
         {
-            if (ResourceExists("volume", name))
+            if (VolumeExists(name))
             {
                 Execute($"volume rm -f {name}");
             }
@@ -83,9 +83,14 @@ namespace Microsoft.DotNet.Docker.Tests
             return $"{ContainerWorkDir}{separator}{relativePath}";
         }
 
-        private static bool ResourceExists(string type, string id)
+        public static bool ImageExists(string tag)
         {
-            ProcessStartInfo startInfo = new ProcessStartInfo("docker", $"{type} ls -q {id}");
+            return ResourceExists("image", tag);
+        }
+
+        private static bool ResourceExists(string type, string filterArg)
+        {
+            ProcessStartInfo startInfo = new ProcessStartInfo("docker", $"{type} ls -q {filterArg}");
             startInfo.RedirectStandardOutput = true;
             Process process = Process.Start(startInfo);
             process.WaitForExit();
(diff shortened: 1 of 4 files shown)


test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -14,9 +14,38 @@ namespace Microsoft.DotNet.Docker.Tests
...
+        private static string OsFilter => Environment.GetEnvironmentVariable("IMAGE_OS_FILTER");
         private static string VersionFilter => Environment.GetEnvironmentVariable("IMAGE_VERSION_FILTER");
 
+        private static ImageDescriptor[] LinuxTestData = new ImageDescriptor[]
+            {
+                new ImageDescriptor { DotNetCoreVersion = "1.0", SdkVersion = "1.1" },
+                new ImageDescriptor { DotNetCoreVersion = "1.1", RuntimeDepsVersion = "1.0" },
+                new ImageDescriptor { DotNetCoreVersion = "2.0" },
+                new ImageDescriptor { DotNetCoreVersion = "2.0", OsVariant = OS.Jessie },
+                new ImageDescriptor { DotNetCoreVersion = "2.0", OsVariant = OS.Stretch, SdkOsVariant = "", Architecture = "arm" },
```
ravimeda: Nit - please check line length.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
```diff
@@ -44,7 +44,7 @@ public void DeleteImage(string tag)
 
         public void DeleteVolume(string name)
         {
-            if (ResourceExists("volume", name))
+            if (VolumeExists(name))
```
ravimeda: I can't see how introducing `VolumeExists` makes things better.  Since this can be achieved with `ResourceExists("volume", $"-f \"name={name}\"")`

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
```diff
@@ -82,7 +68,7 @@ public static IEnumerable<object[]> GetVerifyImagesData()
             }
 
             // Filter out test data that does not match the active architecture and version filters.
-            return testData
+            return (DockerHelper.IsLinuxContainerModeEnabled ? LinuxTestData : WindowsTestData)
```
ravimeda: Neat!

test/Microsoft.DotNet.Docker.Tests/ImageTests.cs
MichaelSimons: This is the max length showable in the github viewer without clipping.  Leaving as is.

test/Microsoft.DotNet.Docker.Tests/DockerHelper.cs
MichaelSimons: I think it is more readable and provides symmetry with the DeleteImage implementation.  Leaving as is.

Issue #529, opened 2018-01-21
Docker assets should be cleaned up towards the end of a CI check


ravimeda:
## Steps to reproduce the issue
1. Submit a dummy PR. To avoid merging the changes, add `[WIP]` in the title.
2. Wait for CI checks to finish
3. Verify if Docker assets are cleaned up towards the end of each CI check

## Expected behavior

All Docker assets that were created while running the test should be cleaned up.  

Running `docker system prune -a -f` would clear up everything including base OS image such as Windows Server Core.  These images are several GiB in size, and are not modified during a CI run. Only time these images need to be re-pulled is when a newer image is available in the registry.  Hence it would be more efficient to avoid deleting such base OS images, and only delete images that are built during each CI.

## Actual behavior

`%ProgramData%\Docker` that contains the images pulled and created during the CI check is not cleaned up.  This can result in filling up the disk, and the test VM becomes unusable.

ravimeda:
Attempt to fix this issue through https://github.com/dotnet/dotnet-docker-nightly/commit/96618736802d654be05d1165eed998583e8f0a68 introduced a regression.  This is because, the statement `docker ps -a -q | ForEach-Object { docker rm -f $_ }` when run inside a container can delete the container itself. Exit code thrown is 137.  Linux CI that runs in a container with _testrunner_ image is broken.  Windows CI is fine since it does not run in a container.

MichaelSimons:
Fixed.

PR #536, opened 2018-01-23
Cleanup Docker assets during CI

ravimeda:
Refer to https://github.com/dotnet/dotnet-docker-nightly/issues/529

This fixes the regression introduced through 9661873.  `docker system prune -a -f` will be used in non-Windows path to cleanup Docker assets.

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -31,6 +31,8 @@ function Invoke-CleanupDocker($ActiveOS)
 }
 
 $(docker version) | % { Write-Host "$_" }
+$activeOS = docker version -f "{{ .Server.Os }}"
+Invoke-CleanupDocker $activeOS
 
 if ($UseImageCache) {
     $optionalDockerBuildArgs = ""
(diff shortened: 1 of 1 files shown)


ravimeda:
@dotnet-bot test ci please

build-and-test.ps1
```diff
@@ -40,6 +42,7 @@ else {
 $manifest = Get-Content "manifest.json" | ConvertFrom-Json
 $manifestRepo = $manifest.Repos[0]
 $activeOS = docker version -f "{{ .Server.Os }}"
+Invoke-CleanupDocker $activeOS
```
MichaelSimons: Consider moving this logic along with the logic to retrieve the activeOS before the manifest is loaded.

ravimeda: Ok. Say it is moved to line 34, then the code look like this -

```powershell
$(docker version) | % { Write-Host "$_" }
$activeOS = docker version -f "{{ .Server.Os }}"
Invoke-CleanupDocker $activeOS
```

`docker version` getting called in subsequent statements makes me uncomfortable. These two calls were present prior to this PR, but makes it more glaring now.

PR #542, opened 2018-01-31
libuv package from edge-repo

oanogin:
"Runtime" image invalid due to the this image (runtime-deps), because it hasn't libuv package installed.

### DIFF
--- a/2.1/runtime-deps/alpine/amd64/Dockerfile
+++ b/2.1/runtime-deps/alpine/amd64/Dockerfile
@@ -16,7 +16,7 @@ RUN apk add --no-cache \
         userspace-rcu \
         zlib \
     && apk -X https://dl-cdn.alpinelinux.org/alpine/edge/main add --no-cache \
-        lttng-ust
+        lttng-ust libuv
 
 # Set the invariant mode since icu_libs isn't included (see https://github.com/dotnet/announcements/issues/20)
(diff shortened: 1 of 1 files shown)


MichaelSimons:
@janvorli - Can you take a look and comment?  libuv wasn't in the original list of Alpine dependencies.

janvorli:
Libuv is not a coreclr runtime dependency, it is a dependency of the ASP.NET Core. But I guess that the docker image needs to support ASP.NET Core too, so the addition makes sense.

MichaelSimons:
@janvorli - thanks for info.  ASP.NET Core has their own [images](https://github.com/aspnet/aspnet-docker) although they haven't yet added preview images for alpine.  My initial reaction is that libuv should be added within an ASP.NET Core specific image to avoid adding unnecessary packages to the core runtime images.

@natemcmaster, @richlander

natemcmaster:
We're tracking aspnetcore support for alpine here: https://github.com/aspnet/Universe/issues/833. It's not yet a platform aspnetcore supports.

MichaelSimons:
@oanogin - thanks for making a PR.  As noted in the discussion, the libuv dependency is something that belongs in the ASP.NET Core images.  This issue should be addressed with aspnet/Universe#833.  Closing PR but feel free to discuss.

PR #548, opened 2018-02-02
Refactor update-dependencies tool

### LINKED ISSUE #476: Environment variables used in update-dependencies are named poorly
RuntimeReleasePrefix is used as the cli release prefix
CliReleasePrefix is used as the Dockerfile version folder the tool runs on.

MichaelSimons:
- Updated to work off of new orchestrated build info.  This eliminates the clunky pattern of needing to retrieve the runtime version directly from the CLI repo.
- Dropped support for everything pre 2.1 which wasn't being used.
- Dockerized the PS runner script.
- Added command line args  over using EnvVars so that it is easier to invoke manually.

There are three TODO comments which are self-explanatory.  I will wait on merging until ProductVersion is available.

Fixes #476


update-dependencies/Invoke-UpdateDependencies.ps1
```diff
@@ -0,0 +1,32 @@
...
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+$imageName = "update-dependencies"
+
+try {
+    & docker build -t $imageName -f $PSScriptRoot\Dockerfile --pull $PSScriptRoot
+    if ($LastExitCode -ne 0) {
```
ravimeda: Not sure of the convention, but `LastExitCode`  should be all uppercase.

update-dependencies/DockerfileShaUpdater.cs
```diff
@@ -29,11 +29,11 @@ public DockerfileShaUpdater(string dockerfilePath) : base()
         }
 
         protected override string TryGetDesiredValue(
-            IEnumerable<DependencyBuildInfo> dependencyBuildInfos,
-            out IEnumerable<BuildInfo> usedBuildInfos)
+            IEnumerable<IDependencyInfo> dependencyBuildInfos,
```
ravimeda: `using Microsoft.DotNet.VersionTools;` that will not be required now.

update-dependencies/Program.cs
```diff
@@ -3,152 +3,138 @@
...
+                    else
+                    {
+                        await CreatePullRequest(updateResults);
                     }
                 }
             }
             catch (Exception e)
             {
                 Console.Error.WriteLine($"Failed to update dependencies:{Environment.NewLine}{e.ToString()}");
+                Environment.Exit(1);
```
ravimeda: Clarification -  please could you clarify why `Environment.Exit(1);` is needed.

update-dependencies/NuGet.config
```diff
@@ -2,5 +2,6 @@
 <configuration>
   <packageSources>
     <add key="dotnet-buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
+    <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/" />
```
dagood: iirc, this url ends up using the v2 NuGet feed. v3 should work better.

update-dependencies/update-dependencies.csproj
```diff
@@ -3,11 +3,13 @@
...
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="1.0.27-prerelease-01428-01" />
-    <PackageReference Include="System.Diagnostics.TextWriterTraceListener" Version="4.0.0" />
-    <PackageReference Include="NuGet.Packaging" Version="4.0.0" />
+    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="2.2.0-prerelease-02431-01" />
+    <PackageReference Include="NuGet.Packaging" Version="4.3.0" />
```
dagood: Can this package ref be removed? This is the version that `VersionTools` depends on (as of https://github.com/dotnet/buildtools/pull/1884).

update-dependencies/Program.cs
```diff
@@ -3,152 +3,138 @@
...
-                    return false;
-                }
-            }
-
-            return true;
+            Environment.Exit(0);
         }
 
-        private static DependencyUpdateResults UpdateFiles()
+        private static async Task<DependencyUpdateResults> UpdateFiles()
```
dagood: `async` methods missing `Async` suffix in names.

update-dependencies/Program.cs
```diff
@@ -3,152 +3,138 @@
...
-            // Adjust the LatestReleaseVersion since it is not the full version and all consumers here need it to be.
-            cliBuildInfo.LatestReleaseVersion = $"{s_config.RuntimeReleasePrefix}-{cliBuildInfo.LatestReleaseVersion}";
-            CliDependencyHelper cliDependencyHelper = new CliDependencyHelper(cliBuildInfo.LatestReleaseVersion);
-            string sharedFrameworkVersion = cliDependencyHelper.GetSharedFrameworkVersion();
-
-            IEnumerable<DependencyBuildInfo> buildInfos = new[]
+            Trace.TraceInformation($"Retrieving build info from '{Options.BuildInfoUrl}'");
+            Stream stream = await (new HttpClient().GetStreamAsync(Options.BuildInfoUrl));
+            XDocument buildInfoXml = XDocument.Load(stream);
+            OrchestratedBuildModel buildInfo = OrchestratedBuildModel.Parse(buildInfoXml.Root);
```
dagood: I was going to ask if `BuildManifestClient.FetchManifestAsync` worked here, but it uses the GitHub API (rather than the `raw` endpoint) so this code is probably worth it to get around the pretty small anonymous rate limit. Might be worth adding functionality to the VersionTools GitHub client itself to derive a `raw` url when fetching something anonymously, actually.

update-dependencies/Program.cs
```diff
@@ -3,152 +3,138 @@
...
-
-            BuildInfo cliBuildInfo = BuildInfo.Get(CliBuildInfoName, s_config.CliVersionUrl, fetchLatestReleaseFile: false);
-            // Adjust the LatestReleaseVersion since it is not the full version and all consumers here need it to be.
-            cliBuildInfo.LatestReleaseVersion = $"{s_config.RuntimeReleasePrefix}-{cliBuildInfo.LatestReleaseVersion}";
-            CliDependencyHelper cliDependencyHelper = new CliDependencyHelper(cliBuildInfo.LatestReleaseVersion);
-            string sharedFrameworkVersion = cliDependencyHelper.GetSharedFrameworkVersion();
-
-            IEnumerable<DependencyBuildInfo> buildInfos = new[]
+            Trace.TraceInformation($"Retrieving build info from '{Options.BuildInfoUrl}'");
+            Stream stream = await (new HttpClient().GetStreamAsync(Options.BuildInfoUrl));
```
dagood: Probably not a big deal in this context, but undisposed client and stream.

update-dependencies/Program.cs
```diff
@@ -3,152 +3,138 @@
...
+            BuildIdentity sdkBuild = buildInfo.Builds
+                .First(build => string.Equals(build.Name, "cli", StringComparison.OrdinalIgnoreCase));
+            BuildIdentity coreSetupBuild = buildInfo.Builds
+                .First(build => string.Equals(build.Name, "core-setup", StringComparison.OrdinalIgnoreCase));
+
+            // TODO:  Remove once ProductVersion is written to build-info
+            EndpointModel endpointInfo = buildInfo.Endpoints
+                .First(endpoint => string.Equals(endpoint.Id, "dotnetcli", StringComparison.OrdinalIgnoreCase));
+            BlobArtifactModel blobInfo = endpointInfo.Artifacts.Blobs
+                .First(artifact => artifact.Id.StartsWith("Runtime", StringComparison.OrdinalIgnoreCase));
```
dagood: Just want to point out that using this endpoint means you want to trigger on `installers.semaphore`, since that indicates the endpoint was published to and added to the manifest.

update-dependencies/DockerfileShaUpdater.cs
MichaelSimons: Removed

update-dependencies/Invoke-UpdateDependencies.ps1
MichaelSimons: This is the convention we have been using across the Docker repos so I would prefer to leave it for consistency.  I don't see this as a widely adopted convention.

update-dependencies/Program.cs
MichaelSimons: There is more than one way this can be achieved but it is a nice way to have the console app exit gracefully with a non-zero exit code.

update-dependencies/NuGet.config
MichaelSimons: Fixed

update-dependencies/Program.cs
MichaelSimons: Renamed

update-dependencies/Program.cs
MichaelSimons: Yep - I that is was I was thinking needed to happen - thanks for confirming.

update-dependencies/update-dependencies.csproj
MichaelSimons: Removed.

update-dependencies/Program.cs
MichaelSimons: I admit was being a little lazy for this console app :)  I fixed it.

MichaelSimons:
> Right now there's a split between master, still creating the old build-infos, and prodcon builds (release/2.1), creating the new build manifest used here. Will this updater only be used on prodcon builds? If not, dotnet/buildtools#1853 could become a blocker for daily docker builds pointing at master.

The POR is to use prodcon build manifest for the 2.1 release.  Once we move off of release/2.1 and start pulling builds from master again, I am hoping the state of master (e.g. the richness of the version info) will more closely align with 2.1.  Any specialization logic to read a different manifest schema can be added at that time and encapsulated into the single place the constructs the IDependencyInfo.

MichaelSimons:
I don't think waiting is necessary since the core-setup ProductVersion is now in place.  The CLI BuildId will work for now.  I will log a followup issue to convert is to ProductVersion once that work is done.

MichaelSimons:
Issue to use CLI ProductVersion once available - https://github.com/dotnet/dotnet-docker-nightly/issues/550

